PR #34, opened 2022-11-08
dart: Deduplicate test data

### LINKED ISSUE #21: Gherkin testdata duplication

Testdata for gherkin are duplicated into each implementation. In case of update of the testdata, it generates huge diffs which are making PRs difficult to review properly (cf. cucumber/common#1741)

Stop duplicating testdata and refer to those using relative paths
Or at least duplicating the testdata during the build, but ignoring those from the repo

### 📚 Any additional context?

Why the purpose of the duplication? Is it because of the subrepos? 

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/developer_experience.md)*

mpkorstanje:
Fixes #21 .


mpkorstanje:
Can't be fixed. Darts parser does not retain the `StepKeywordType`.

Issue #36, opened 2022-11-08
javascript: Acceptance tests do not include tokens


mpkorstanje:
### 🤔 What's the problem you've observed?

The acceptance tests for Javascript currently do *not* verify if the parser produces the right tokens:

https://github.com/cucumber/gherkin/blob/1cc2ef0153603c5e51818b59ff6d632238491226/javascript/Makefile#L13-L16
https://github.com/cucumber/gherkin/blob/1cc2ef0153603c5e51818b59ff6d632238491226/java/Makefile#L42

Compare with:

https://github.com/cucumber/gherkin/blob/1cc2ef0153603c5e51818b59ff6d632238491226/java/Makefile#L13-L17
https://github.com/cucumber/gherkin/blob/1cc2ef0153603c5e51818b59ff6d632238491226/java/Makefile#L42

This is a problem because:

> As of May 2016 Gherkin is implemented in 8 languages. This number is likely to increase to a dozen within a year or two. Very few programmers are intimate with all those languages. Therefore, in order to make it as easy as possible to refactor, fix bugs, add features and release packages it is essential that all implementations have a similar structure.

_[CONTRIBUTING.md - consistency between implementations](https://github.com/cucumber/gherkin/blob/main/CONTRIBUTING.md#consistency-between-implementations)_

### ✨ Do you have a proposal for making it better?

Include the tokens in the acceptance tests again.

### 📚 Any additional context?

The tokens were removed with https://github.com/cucumber/common/pull/448 and never put back once reverted.

mpkorstanje:
@aslakhellesoy can you pick this up?

PR #37, opened 2022-11-08
python: Enable markdown acceptance tests

mpkorstanje:
Follow up for https://github.com/cucumber/common/pull/2103

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -8,8 +8,8 @@ SOURCE_FILES = $(shell find . -name "*.py" | grep -v $(GHERKIN_PARSER))
 GHERKIN = bin/gherkin
 GHERKIN_GENERATE_TOKENS = bin/gherkin-generate-tokens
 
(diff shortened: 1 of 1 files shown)


mpkorstanje:
@temyers looks like we missed a few things when enabling the acceptance tests.

You will at the very least have to fix `gherkin_generate_tokens.py` to handle markdown files. And you may also have to generate some acceptance tests data. I was working on the assumption that the Javascript implementation had included these already (I was wrong, https://github.com/cucumber/gherkin/issues/36).

temyers:
Dang,  ok I'll try and have a look.  

Interesting that this wasn't picked up during the original merge?  looking at the referenced issue, they were disabled?

temyers:
@mpkorstanje Please see: #64
I've started making changes there.  Still WIP.

Worth moving any discussion under that PR?

mpkorstanje:
I am happy to let it supercede this one.

PR #44, opened 2022-11-09
Fix Georgian localization

mpkorstanje:
### 🤔 What's changed?
In https://github.com/cucumber/common/pull/2041 only half a unicode glyph was pasted to create `Მოცემულია `. This half glyph breaks code generation in other projects, e.g:

https://github.com/cucumber/cucumber-jvm/actions/runs/3429109517/jobs/5714282373

This presumably (based on some searching) should have been `მოცემულია`.Though I don't speak Georgian. Please send another pull request if this is not correct.

### 📋 Checklist:

  - [x] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../CHANGELOG.md), linking to this pull request.


avsd:
> @avsd Did I guess this translation correctly? ☝️

Visually it looks exactly the same! I'm wondering how could I have lost half a glyph :D - and how could it have been rendered correctly with half a glyph lost? Sorry for that!

avsd:
@mpkorstanje, ok, I know the answer. It contained CAPITAL letter (not used in modern Georgian):

https://www.compart.com/en/unicode/U+1C9B

instead of a correct SMALL letter:

https://www.compart.com/en/unicode/U+10DB

Possible reason why it might have happened is capitalization, which is probably performed by some script when generating the files. If that's the case, it may happen again

image




It might have been auto-capitalization

avsd:
ok, apparently it was introduced by me, during the very initial commit: https://github.com/cucumber/common/pull/2041/commits/7054aa522e112d020ce521ee80b87786080ebb49

Maybe because I was using some unusual IDE (either Mac's TextEdit or GitHub's built-in editor) that did auto-capitalization after hitting "space". But the problem is that this incorrect "Georgian Capital Letter M" might have been spread to numerous repositories, all of them requiring a fix. What's the best way to do it, @mpkorstanje ?

mpkorstanje:
I don't think anything has to be done.

I've just extracted `gherkin` from the `common` repository. Which is why the fix happened here.

The mistake is duplicated in a few other repositories but those are subtrees of `common`. The mistake there isn't a problem because they're not part of any release.

What has been released will have a patch version to fix it.

avsd:
Ok, good! :) In that case, here are two more locations this letter appeared: https://github.com/cucumber/gherkin/pull/45

PR #46, opened 2022-11-10
PHP: Fix generated code formatting

### LINKED ISSUE #35: php: Parser is not generated with the right formatting

One of the goals of setting up the workflow for this project was that generating parsers should only depend on Berp. This simplifies the image needed to build this project. However PHP also formatted the parser using vendored tools. This effectively adds a dependency on PHP. 

https://github.com/cucumber/gherkin/blob/6b84ba8e92c901192c707ac66ee61695cbf1b9f3/php/Makefile#L47-L52

Either
 * fix `gherkin-php.razor` to produce correctly formatted files
 * or ignore the problem. It is generated code.

### :computer: Additional context

In https://github.com/cucumber/gherkin/pull/3 I've removed the code formatter and fixed most of the razor file. I couldn't fix everything.

ciaranmcnulty:
Closes #35

Adds `src-generated` folder to the coding style check, and fix the razor template so it passes

### DIFF
--- a/php/.php-cs-fixer.php
+++ b/php/.php-cs-fixer.php
@@ -2,6 +2,7 @@
 $finder = PhpCsFixer\Finder::create()
     ->in([
         __DIR__ . '/src',
+        __DIR__ . '/src-generated',
         __DIR__ . '/tests',
         __DIR__ . '/bin',
     ])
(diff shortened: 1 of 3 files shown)


mpkorstanje:
Cheers! Much appreciated!

PR #52, opened 2022-11-11
Draft: [PHP] Introduce Mutation testing

ciaranmcnulty:
This introduces mutation testing for the non-generated code; Draft for now as I want to close off some of the detected mutants

In order to be able to use the acceptance tests with the mutation testing tools, the acceptance testing has been transposed into PHPUnit rather than the existing bash/make version

The `gherkin-generate-tokens` binary is removed as that was only for testing, but the `bin/gherkin` is left the same


ciaranmcnulty:
I've cleaned out a lot of the 'fixes' as they were cluttering the PR

It looks like the PHP composer install is being rate-limited at GitHub, which I don't think should happen in GHA! Investigating...

ciaranmcnulty:
ok, #55 might fix the rate limits + I'll rebase on top of that when it's merged

php/Makefile
```diff
@@ -39,12 +29,13 @@ clean: ## Remove all build artifacts and files generated by the acceptance tests
 
 .DELETE_ON_ERROR:
 
-acceptance: .built $(TOKENS) $(ASTS) $(PICKLES) $(ERRORS) $(SOURCES) ## Build acceptance test dir and compare results with reference
+acceptance: .built  ## Test parser against test data
```
ciaranmcnulty, line 32: Not sure if this should be `.PHONY` now?

.github/workflows/test-php.yml
```diff
@@ -56,8 +57,23 @@ jobs:
         run: |
           vendor/bin/php-cs-fixer --dry-run --diff fix
           vendor/bin/psalm --no-cache
-          vendor/bin/phpunit
+          vendor/bin/phpunit --testsuite unit
```
mpkorstanje, line 60: We can run all the tests in a regular run. The separation between acceptance and unit existed only because we had different methods for testing.

php/Makefile
mpkorstanje: You can remove the goal entirety.

Do update the `php/README.md` with info about the structure.

PR #55, opened 2022-11-13
PHP - Add github auth token to composer configuration

ciaranmcnulty:
This writes the github auth token to a config file in ~/.composer, which then uses it when downloading packages

This will help with github rate limiting, as was happening on https://github.com/cucumber/gherkin/pull/52

Documentation here: https://github.com/shivammathur/setup-php#github-composer-authentication

### DIFF
--- a/.github/workflows/test-php.yml
+++ b/.github/workflows/test-php.yml
@@ -10,6 +10,9 @@ on:
       - main
   workflow_call:
 
+permissions:
+  contents: read
+
 jobs:
   test-php:
     runs-on: ubuntu-latest
--- a/.github/workflows/test-php.yml
+++ b/.github/workflows/test-php.yml
@@ -25,6 +25,8 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php }}"
+        env:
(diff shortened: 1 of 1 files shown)


mpkorstanje (approved):
LGTM. What about the other projects like messages, cucumber expressions, ect?

.github/workflows/test-php.yml
```diff
@@ -25,6 +25,8 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php }}"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mpkorstanje, line 29: This may need a permission reduction though, tokens don't come with a sensible minimal set of permissions. Something like:

```
permissions:
      contents: read
```

mpkorstanje (changes requested):
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

ciaranmcnulty:
@mpkorstanje doesn't seem to have broken anything 👍 

Messages is the only other project using composer right now

PR #72, opened 2022-12-17
Erlang: Upgrade OTP

mpkorstanje:
The builds now run on Ubuntu 22.04. This means that OTP 22.2 is no longer available.

See:

https://github.com/erlef/setup-beam#compatibility-between-operating-system-and-erlangotp


mpkorstanje:
@WannesFransen1994 could you help out with this one? It looks like the Erlang OTP versions are quite complicated and I couldn't immediately figure out which version to use.

I've disabled the Elixir builds until then.

Issue #82, opened 2022-12-25
The minimum messages version should still be 19.1.4


mpkorstanje:
>  @mattwynne I do think it that the version just isn't declared correctly in the Gemfile. The minimum should still be 19.1.4.  So using `~>` is probably the wrong operator. We want multiple majors.
>
> Looks like only Java got this right
> 
> ```
> <artifactId>messages</artifactId>
> <version>[19.1.4,22.0.0)</version>
> ```

_Originally posted by @mpkorstanje in https://github.com/cucumber/gherkin/issues/81#issuecomment-1364601964_

mpkorstanje:
Perhaps we need to use the `widen` strategy.

https://docs.renovatebot.com/configuration-options/#rangestrategy

mpkorstanje:
This also applies to the HTML formatter.

Note: This may get confusing because NodeJS also has the `peerDependency` version.

PR #83, opened 2022-12-27
Allow messages v19.1.4 to v20.0.1 to be used

mpkorstanje:
### 🤔 What's changed?
Allow messages v19.1.4 to v20.0.1 to be used by Javascript, PHP and Ruby.

With v19 [the messages protocol was expanded with keyword types][1]. Since then no relevant changes for Gherkin have been made. This means we can expand the range of allowed message versions somewhat.

1: https://github.com/cucumber/messages/blob/main/CHANGELOG.md#1900---2022-05-31

The documentation for the syntax of each can be found here:
 - https://semver.npmjs.com/
 - https://getcomposer.org/doc/articles/versions.md
 - https://guides.rubygems.org/patterns/

Fixes: https://github.com/cucumber/gherkin/issues/82

### 📋 Checklist:

- [X] Users should know about my change
  - [X] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../CHANGELOG.md), linking to this pull request.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org).
 This document is formatted according to the principles of [Keep A CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+- [Javascript, PHP, Ruby] Allow messages v19.1.4 to v20.0.1 to be used
 
 ## [26.0.2] - 2022-12-27
 ### Fixed
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -40,7 +40,7 @@
     "typescript": "4.9.4"
   },
   "dependencies": {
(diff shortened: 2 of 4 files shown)


mpkorstanje:
@mattwynne  see also https://github.com/cucumber/html-formatter/pull/206

Though there we are using `peerDependencies` for Javascript.  Shouldn't we do the same here?

davidjgoss (approved):

mpkorstanje:
@davidjgoss what about the peer depencies?

davidjgoss:
Oh sorry!

For the html formatter it made sense for messages to be a peer dep because it’s used in the context of a cucumber implementation.

Whereas usages of Gherkin can be more varied eg pretty printers, linters so making it a peer dep would add some overhead to those use cases.

Maybe run with status quo for now and we can review this more broadly across the libraries?

PR #89, opened 2023-01-18
ruby: introduce Gherkin::Query#parent_locations.

botandrose:
### 🤔 What's changed?
Introduces a new Gherkin::Query#parent_locations method in the Ruby implementation.

Cucumber::Core::Compiler will use this method to tell Cucumber::Core::Test::Case the locations of its parent lines, such as `Feature:`, `Background:`, and `Rule:`, so that it will correctly match them during location filtering. This particular strategy was identified as potentially reasonable while pairing with @mattwynne.

This is one part of a two-repo PR to resolve the last remaining bits of the regression described in https://github.com/cucumber/cucumber-ruby/issues/1469

### 📋 Checklist:

- [?] My change requires a change to the documentation.
- [?] Users should know about my change
  - [ ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../CHANGELOG.md), linking to this pull request.

I'm not sure about these last two, since this could be considered as internals from a high-level cucumber point-of-view, but a public-interface addition from the gem point-of-view. Thoughts?

### DIFF
--- a/ruby/lib/gherkin/query.rb
+++ b/ruby/lib/gherkin/query.rb
@@ -2,10 +2,17 @@ module Gherkin
   class Query
     def initialize
       @ast_node_locations = {}
+      @scenario_parent_locations = {}
+      @background_locations = {}
     end
 
     def update(message)
       update_feature(message.gherkin_document.feature) if message.gherkin_document
+    end
+
+    def scenario_parent_locations(scenario_node_id)
+      return @scenario_parent_locations[scenario_node_id] if @scenario_parent_locations.has_key?(scenario_node_id)
+      raise AstNodeNotLocatedException, "No scenario parent locations found for #{scenario_node_id} }. Known: #{@scenario_parent_locations.keys}"
     end
 
     def location(ast_node_id)
@@ -20,27 +27,27 @@ module Gherkin
       store_nodes_location(feature.tags)
 
       feature.children.each do |child|
-        update_rule(child.rule) if child.rule
-        update_background(child.background) if child.background
-        update_scenario(child.scenario) if child.scenario
+        update_rule(feature, child.rule) if child.rule
+        update_background(feature, child.background) if child.background
+        update_scenario(feature, child.rule, child.scenario) if child.scenario
       end
     end
 
-    def update_rule(rule)
+    def update_rule(feature, rule)
       return if rule.nil?
       store_nodes_location(rule.tags)
-
       rule.children.each do |child|
-        update_background(child.background) if child.background
-        update_scenario(child.scenario) if child.scenario
+        update_background(rule, child.background) if child.background
+        update_scenario(feature, rule, child.scenario) if child.scenario
       end
     end
 
-    def update_background(background)
+    def update_background(parent, background)
       update_steps(background.steps)
+      @background_locations[parent] = background.location
     end
 
-    def update_scenario(scenario)
+    def update_scenario(feature, rule, scenario)
       store_node_location(scenario)
       store_nodes_location(scenario.tags)
       update_steps(scenario.steps)
@@ -48,6 +55,13 @@ module Gherkin
         store_nodes_location(examples.tags || [])
         store_nodes_location(examples.table_body || [])
       end
+
+      @scenario_parent_locations[scenario.id] = [
+        feature.location,
+        @background_locations[feature],
+        rule&.location,
+        @background_locations[rule],
+      ].compact
     end
 
     def update_steps(steps)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 - (i18n) Added Danish translation of "Rule"
 - (i18n) Added Dutch translation of "Rule"
 - (i18n) Added Esperanto translation of "Rule"
+- [Ruby] Added `Gherkin::Query#parent_locations` for determining a scenario's parents' line numbers ([#89](https://github.com/cucumber/gherkin/pull/89))
 
 ## [26.2.0] - 2023-04-07
 ### Changed
--- a/ruby/spec/gherkin/query_spec.rb
+++ b/ruby/spec/gherkin/query_spec.rb
@@ -47,6 +47,7 @@ describe Gherkin::Query do
         Examples:
           | Status |
           | passed |
+          | failed |
 
       @rule-tag
       Rule: this is a rule
@@ -68,6 +69,56 @@ describe Gherkin::Query do
           messages.each { |message| subject.update(message) }
         end.not_to raise_exception
       end
+    end
+  end
+
+  describe '#scenario_parent_locations' do
+    before do
+      messages.each { |message| subject.update(message) }
+    end
+
+    let(:background) { find_message_by_attribute(gherkin_document.feature.children, :background) }
+    let(:scenarios) { filter_messages_by_attribute(gherkin_document.feature.children, :scenario) }
+
+    context 'without rule' do
+      let(:scenario) { scenarios.first }
+
+      it 'provides the feature and background locations of a given scenario node id' do
+        expect(subject.scenario_parent_locations(scenario.id)).to eq([
+          gherkin_document.feature.location,
+          background.location,
+        ])
+      end
+    end
+
+    context 'with rule' do
+      let(:rule) { find_message_by_attribute(gherkin_document.feature.children, :rule) }
+      let(:rule_background) { find_message_by_attribute(rule.children, :background) }
+      let(:scenario) { find_message_by_attribute(rule.children, :scenario) }
+
+      it 'provides the feature, background, rule, and rule background locations of a given scenario node id' do
+        expect(subject.scenario_parent_locations(scenario.id)).to eq([
+          gherkin_document.feature.location,
+          background.location,
+          rule.location,
+          rule_background.location,
+        ])
+      end
+    end
+
+    context 'in a scenario outline' do
+      let(:scenario) { scenarios.last }
+
+      it 'provides the feature and background locations of a given scenario outline node id' do
+        expect(subject.scenario_parent_locations(scenario.id)).to eq([
+          gherkin_document.feature.location,
+          background.location,
+        ])
+      end
+    end
+
+    it 'raises an exception if called with an invalid id' do
+      expect { subject.scenario_parent_locations("BAD") }.to raise_error(Gherkin::AstNodeNotLocatedException)
     end
   end
 


ruby/lib/gherkin/query.rb
```diff
@@ -20,34 +27,42 @@ def update_feature(feature)
       store_nodes_location(feature.tags)
 
       feature.children.each do |child|
-        update_rule(child.rule) if child.rule
-        update_background(child.background) if child.background
-        update_scenario(child.scenario) if child.scenario
+        update_rule(feature, child.rule) if child.rule
+        update_background(feature, child.background) if child.background
+        update_scenario(feature, nil, child.scenario) if child.scenario
```
luke-hill, line 32: 5second check. Should this be `child&rule` to cover where the child does / doesn't respond to rule?

luke-hill: Right, so this should be added then right?

botandrose: Hmm, I'm not following. Are you saying that we might want to change line 30 by adding a safe navigation operator (&.) to the postfix guard clause, like so?

`update_rule(feature, child.rule) if child&.rule`

luke-hill: No the line I've commented on here - 32. you have a `nil` rule. this feels wrong?

botandrose: @luke-hill Ahh, I gotcha now. Let me dig in and see if I can figure out why I thought `nil` made sense back when I wrote this.

botandrose: @luke-hill I really can't figure out why I put `nil` there. All the tests still pass when replacing it with `child.rule`. No need for safe nav operator I don't think, since child always responds to `#rule` (it just might be nil, which is handled in the called method). I'll push a commit.

botandrose: @luke-hill Okay, pushed. Could probably do with a squash when merging.

luke-hill:
So I've had another quick scan of this @botandrose - I think firstly this probably "feels" like a major change, because you're refactoring a huge chunk of how features e.t.c. are updated (I know they're private), but this entire area is a private DSL essentially.

Looking at the code changes (Not spec), I actually think it all makes sense and is fine. Next step would be to review specs.

luke-hill:
@botandrose so I'm keeping you in the loop and working "semi" linearly. Are we saying this should / must go in first and be released before thinking about other fixes?

ruby/lib/gherkin/query.rb
```diff
@@ -20,34 +27,42 @@ def update_feature(feature)
...
       rule.children.each do |child|
-        update_background(child.background) if child.background
-        update_scenario(child.scenario) if child.scenario
+        update_background(rule, child.background) if child.background
+        update_scenario(feature, rule, child.scenario) if child.scenario
       end
     end
 
-    def update_background(background)
+    def update_background(parent, background)
```
luke-hill, line 45: Think this should be `rule` based on how you've done it in the other methods.

ruby/lib/gherkin/query.rb
```diff
@@ -20,34 +27,42 @@ def update_feature(feature)
...
       update_steps(background.steps)
+      parent_id = parent.respond_to?(:id) ? parent.id : parent.object_id # Cucumber::Messages::Feature doesn't have an id :(
```
luke-hill, line 47: Then I'm assuming this might need tweaking?

ruby/lib/gherkin/query.rb
```diff
@@ -20,34 +27,42 @@ def update_feature(feature)
...
+    def update_scenario(feature, rule, scenario)
       store_node_location(scenario)
       store_nodes_location(scenario.tags)
       update_steps(scenario.steps)
       scenario.examples.each do |examples|
         store_nodes_location(examples.tags || [])
         store_nodes_location(examples.table_body || [])
       end
+
+      @scenario_parent_locations[scenario.id] = [
```
luke-hill, line 60: To save time, as this is the only bit of the new API that's completely new, can you explain it in a line or two of prose so I don't need to dig into the whole API surface for gherkin

ruby/lib/gherkin/query.rb
```diff
@@ -20,34 +27,42 @@ def update_feature(feature)
...
+        feature.location,
+        @background_scopes[feature.object_id]&.location,
```
luke-hill, line 62: can the safe navigation operators here ever return nils i.e. be needed?

luke-hill (commented):
I think by and large this might be the last review I do on this, and just go off a bit of faith as it's still a bit alien to me.

botandrose:
@luke-hill Yes, my cucumber-ruby-core PR is built on-top of this new API, so this has to land before that one can take advantage of it. And they are essentially a pair... without one, the other doesn't really make any sense.

ruby/lib/gherkin/query.rb
botandrose: This is the more abstract `parent`, because sometimes the direct parent of a Background is a Feature, and other times its a Rule. For example:
```
Feature: Cool feature is cool
Scenario: I become cool by using cool feature
```

```
Feature: Cool feature is cool
Rule: Only uncool people can become cool
Scenario: Already cool person cannot use cool feature
```

ruby/lib/gherkin/query.rb
botandrose: IMO, the "correct" thing to do here would be to add a `Cucumber::Messages::Feature#id` method, to make it consistent with the rest of the message types, and the we could get rid of this workaround. It seems like an obvious hole to me. But it also seemed out of scope for this PR.

It seems to me that we have some options:
1. I can push another commit adding `C::M::F#id`
2. I can do a follow-up PR cleaning this up

I'm leaning towards 2 to avoid scope creep, but you're the boss! What do you think?

ruby/lib/gherkin/query.rb
botandrose: Sure, this basically is the meat of the entire effort to fix the long-standing Cucumber 4.0 regression https://github.com/cucumber/cucumber-ruby/issues/1469

This block of code is collating all the line numbers of all the parents of the given scenario:
1. The Feature (and its optional description lines)
2. The optional Background and its steps for the _Scenario_
3. The optional Rule line
4. The optional Background and its steps for the _Rule_

It stores these line numbers data structure -- keyed to the scenario -- for later lookup.

ruby/lib/gherkin/query.rb
botandrose: Yeah, the first item is required, but the latter three are optional. In other words, a Scenario always belongs to a Feature (line 1), but may or may not belong to a Background (line 2), a Rule (line 3), and/or a Rule Background (line 4), so we want to handle the nil case for the last three lines.

ruby/lib/gherkin/query.rb
botandrose: @luke-hill Oh derp, 1 is impossible because its a different repo.

botandrose: @luke-hill I guess option 1 would be opening a PR on https://github.com/cucumber/messages , and waiting for that to be merged and released, and then we could merge and release this with a bumped requirement on that version, and then finally merge and release the cucumber-ruby-core PR with a bumped requirement on this.

botandrose: @luke-hill Just checked with crossed fingers, but `C::M::Feature` is still missing `#id` on current master : https://github.com/cucumber/messages/blob/main/ruby/lib/cucumber/messages.deserializers.rb#L244

botandrose: @luke-hill Ugh, and it looks like that code is all auto-generated. Adding an id to feature would change the data structure for ALL implementations of cucumber, not just the ruby one.

Maybe the simplest method would be to just use `#object_id` as the internal data structure key for both Feature and Rule. Or maybe just the object itself!

ruby/lib/gherkin/query.rb
botandrose: Again, as a curiosity, you can see that we're keying the rule by its `#id`, but we have to use `#object_id` for the feature. :(

ruby/lib/gherkin/query.rb
botandrose: @luke-hill Okay, I just pushed a commit that uses the object itself as the key instead, which ends up simplifying everything greatly, including removing the need for so many safe navigation operators. I'm happy with this if you are!

ruby/lib/gherkin/query.rb
luke-hill: Could you dump an example structure. I'm anticipating something like (With dodgy linespacing delib)

```
1 - Feature: Foo
2 - 
3 - Desc of feature
4 - 
5 - Background:
6 -  Given background one
7 - Scenario: Bar
8 - Given baz
9 - Then foo
```

Produces

```
{ feature: [1,3], background: [5], Rule: [], RuleBackground: []}
```

luke-hill: And for my benefit. We're only looking at **parents** here because `locations` returns everything from the scenario down right?

botandrose: @luke-hill Sure, I just changed the `expect` to `puts(...).inspect` in `spec/gherkin/query_spec.rb#L101`, which tests the method with the "Scenario: a ruled scenario" scenario and this is the result (I added line numbers here for easier output interpretting):

```gherkin
1
2    @feature-tag
3    Feature: my feature
4
5      Background:
6        Given a passed background step
7
8      @scenario-tag
9      Scenario: my scenario
10        Given a passed step
11
12      Scenario Outline: with examples
13        Given a <Status> step
14
15        @examples-tag
16        Examples:
17          | Status |
18          | passed |
19          | failed |
20
21      @rule-tag
22      Rule: this is a rule
23        Background:
24          Given the passed step in the rule background
25
26        @ruled-scenario-tag
27        Scenario: a ruled scenario
28          Given a step in the ruled scenario
```

```ruby
[#<Cucumber::Messages::Location:0x00007f4b2805a620 @line=3, @column=5>,
#<Cucumber::Messages::Location:0x00007f4b28158ea0 @line=5, @column=7>,
#<Cucumber::Messages::Location:0x00007f4b2805bd68 @line=22, @column=7>,
#<Cucumber::Messages::Location:0x00007f4b2804efc8 @line=23, @column=9>]
```

To answer your second question, you might notice that some lines are missing, like tag lines (2, 21, 26) and step lines (6, 24, 28). These lines are already matched correctly in `Cucumber::Core::Compiler#source_lines_for_pickle`. The companion PR over there adds a `#parent_source_lines_for_pickle` that uses this PR's `Gherkin::Query#scenario_parent_locations` method to match the last remaining missing pieces, the Feature, the Background, the Rule, and the Rule Background` lines, which is what you see in the output here.

I first looked at extending `#source_lines_for_pickle` to return everything, but other parts of the cucumber internals were relying on only the source lines being returned, not these extra parent lines, so we needed to keep them separate in the `C::C::TestCase` data structure. Hence, this additional method.

luke-hill: Absolutely perfectly explained.

I personally dislike pos array returning bits over a hash. But I'm 99.9% certain you're going to say the other methods work the same way. i.e. it's magic that the first item is the feature ref

luke-hill:
Approved CI to run. Think we're definitely 90% of the way there now. Sorry for the delay, I'm coming back after a long time off.

ruby/lib/gherkin/query.rb
botandrose: I agree with you on that one, but the good news is that we don't actually care about where the lines come from, only that they match the given scenario. I.e. we could append a #shuffle to the array before returning it, and it wouldn't affect a thing.

botandrose:
@luke-hill all good! I really appreciate your time on this and my other contributions throughout the years. Cucumber is my most used testing framework, as well, so again, your time and effort are very much appreciated

luke-hill (commented):
Can you add a changelog entry for this, we'll look into getting this approved and merged shortly.

botandrose:
@luke-hill Okay, added a changelog entry, rebased on current main, squashed, force-pushed. I think this is ready for merge!

luke-hill (approved):
Sorry for delays.

aslakhellesoy:
Hi @botandrose,

Thanks for your making your first contribution to Cucumber, and welcome to the Cucumber committers team! You can now push directly to this repo and all other repos under the cucumber organization! 🍾

In return for this generous offer we hope you will:

* ✅ Continue to use branches and pull requests. When someone on the core team approves a pull request (yours or someone else's), you're welcome to merge it yourself.
* 💚 Commit to setting a good example by following and upholding our [code of conduct](https://cucumber.io/conduct) in your interactions with other collaborators and users.
* 💬 Join the [community Slack channel](http://cucumber.io/support#slack) to meet the rest of the team and make yourself at home.
* ℹ️ Don't feel obliged to help, just do what you can if you have the time and the energy.
* 🙋 Ask if you need anything. We're looking for feedback about how to make the project more welcoming, so please tell us!

On behalf of the Cucumber core team,
Aslak Hellesøy
Creator of Cucumber

luke-hill:
This will go into the next major (v27), eta a week or so.

botandrose:
@luke-hill @mattwynne thanks to both of you for your help on this! When v27 is out the door I'll rebase the other PR.

PR #145, opened 2023-08-11
Test on oldest supported Perl too

ehuelsmann:
### 🤔 What's changed?

Since we support older Perls, test with the oldest supported Perl too.

Captures issues before cpantesters will (although it's good to have those as backup).

:bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

I hope 1 more matrix variant isn't too much of a burden.

### 📋 Checklist:

### DIFF
--- a/.github/workflows/test-perl.yml
+++ b/.github/workflows/test-perl.yml
@@ -13,14 +13,21 @@ on:
 jobs:
   test-perl:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


mpkorstanje (approved):
LGTM.

Do keep the size of test matrix at 2 though. Otherwise the project will take ages to build. 

There are some matrices for other languages in other projects that are larger, but those are not aspirational.

Issue #31, opened 2022-11-08
perl: Acceptance tests do not include parse errors


mpkorstanje:
### 🤔 What's the problem you've observed?

Currently the error set of tokens is not included in the acceptance tests.

https://github.com/cucumber/gherkin/blob/1cc2ef0153603c5e51818b59ff6d632238491226/perl/Makefile#L46

This is a problem because:

> As of May 2016 Gherkin is implemented in 8 languages. This number is likely to increase to a dozen within a year or two. Very few programmers are intimate with all those languages. Therefore, in order to make it as easy as possible to refactor, fix bugs, add features and release packages it is essential that all implementations have a similar structure.

_[CONTRIBUTING.md - consistency between implementations](https://github.com/cucumber/gherkin/blob/main/CONTRIBUTING.md#consistency-between-implementations)_

### ✨ Do you have a proposal for making it better?

Fix the acceptance tests to include the parse errors.

ehuelsmann:
yep. I'll have a look in the morning.

mpkorstanje:
Cheers! No rush at all, it's been like this for a while.

PR #149, opened 2023-08-15
Perl harmonized errors

ehuelsmann:
### 🤔 What's changed?

Test suite for parse errors is now run too.

Consistency. The rest is explained in #31.

:bug: Bug fix (non-breaking change which fixes a defect)

I think this is a minor version at most: the only difference in behaviour is the error code path, which now results in Cucumber Messages (in most cases) instead of exceptions to be caught.

The fact that the exceptions don't auto-stringify anymore doesn't look like a major issue, because they used to be immediately stringified when being thrown (throwing and having a stringification method turn out to be incompatible).

### 📋 Checklist:


ehuelsmann:
Changelog still to come, obviously.

ehuelsmann:
After consulting with @mpkorstanje, he'll review in hindsight, if he wants to.

PR #151, opened 2023-08-16
Check behavior when a rule has a tag with whitespace

brasmusson:
Use CI to explore the behavior with respect to whitespace in tags. Not inteded to be merged.


TheKayneGame:
If it's not intended to be merge, can this PR be closed now?

PR #154, opened 2023-08-20
Add docker container to generate parsers and languages

ehuelsmann:
### 🤔 What's changed?

A dockerized version of Berp.

The `cucumber/build` container is huge. Most people only need it to run Berp. So, instead of downloading and maintaining a multi-gigabyte container, lets use a lean one with exactly one purpose.

This container likely causes permissions conflicts: the `cukebot` user has hard-coded UID 1000 whereas not everybody will be running as that user. The release container had trickery to be able to run "as the current user". Is it easy to transplant and do we want/need that? (was it even in the container? Or was it in the makefile before?)

### 📋 Checklist:


docker-berp/Dockerfile
```diff
@@ -0,0 +1,70 @@
+# Builds a docker image used for building most projects in this repo. It's
+# used both by contributors and CI.
+#
+FROM ubuntu:22.04
```
luke-hill, line 4: Would it not be better to depend on some dotnet variant that is supported / good?

ehuelsmann: Yes. I think that makes sense. I'll check what I can do about this.

ehuelsmann: @luke-hill , I've now taken the dotnet/sdk base image instead of installing dotNet itself. I've tried using just the runtime, but that turned out not to be "enough" for our purpose (doesn't include some tool we need).

ehuelsmann:
Branch rebased (explaining the force-push).

docker-berp/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+FROM mcr.microsoft.com/dotnet/sdk:6.0
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update \
+    && apt-get install --assume-yes \
+    locales
+
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
```
luke-hill, line 13: Usually this is the only lang env variable needed for things like system locale.

i.e. when using ruby and doing `Time.now` this pulls in from this. Are the others needed for something else?

docker-berp/Dockerfile
```diff
@@ -0,0 +1,51 @@
+# Builds a docker image used for building most projects in this repo. It's
+# used both by contributors and CI.
+#
+FROM mcr.microsoft.com/dotnet/sdk:6.0
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update \
+    && apt-get install --assume-yes \
+    locales
```
luke-hill, line 10: I've used this package before and its primary use is switching/checking things on a CLI, i.e. timezone

Are we needing to do this? Or would just using an interactive console / script suffice?

docker-berp/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    --shell /bin/bash \
+    "$USER"
+
+## Trigger first run experience by running arbitrary cmd to populate local package cache
+RUN dotnet --list-sdks
+
+USER $USER
+
+## As a user install node and npm via node version-manager
+WORKDIR /home/$USER
```
luke-hill, line 42: Should this be doing something like `apt-get install nodejs` or an equiv?

luke-hill (commented):
Maybe irrelevant thoughts. But thoughts that might help

docker-berp/Dockerfile
ehuelsmann: No need: this container won't contain an entire development environment anymore. It will just contain enough to run Berp for all languages to generate the parser.

docker-berp/Dockerfile
ehuelsmann: Yes: dotNet expects UTF8 locales to be installed. There's no need to change it on the cli, but dotNet needs it for its operation.

docker-berp/Dockerfile
ehuelsmann: The order of priority is:

1. LANGUAGE
2. LC_ALL
3. LC_xxx, while taking into account the locale category
4. LANG

And I ran into problems setting LANG. So I set the other two too. (I didn't know the order of priority then. Looked that up to answer your question...)

luke-hill (approved):
Get other reviews

mpkorstanje:
Cheers @ehuelsmann ! Took me a while to get to this but this is actually really good improvement. A belated thanks for kicking it off.

PR #169, opened 2023-09-07
Check PHP has populated $argv

ciaranmcnulty:
This will always be a `non-empty-list<string>` in a CLI script but because it might be `null` in other contexts Psalm is strict about checking for it here


ciaranmcnulty:
(I couldn't push to the branch)

mpkorstanje:
Ah. The commiters group had fallen of the project. Should be fixed now.

mpkorstanje:
I've pushed your changes to #168

PR #170, opened 2023-09-08
i18n: Add 'regel' for 'rule' in Dutch

Coedice:
Added a translation of "rule" in Dutch: "regel"

Resolves: https://github.com/cucumber/gherkin/issues/105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 - (i18n) Added Malayalam localization
 - (i18n) Added 'ed' to Italian ([#31](https://github.com/cucumber/gherkin/issues/160))
 - (i18n) Added Danish translation of "Rule"
+- (i18n) Added Dutch translation of "Rule"
 
 ## [26.2.0] - 2023-04-07
 ### Changed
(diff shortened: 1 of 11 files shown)


luke-hill:
Can you add a changelog entry @Coedice

luke-hill:
> I didn't see this change mentioned in the change log at https://github.com/cucumber/cucumber-js/blob/main/CHANGELOG.md

So you wouldn't see a dutch changelog entry there @smuller-nl however you could/should see an update similar to this one: https://github.com/cucumber/cucumber-ruby/commit/b6779f377982975c412244c5ca653d9da3629cd8#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR15

> And the Localization documentation at https://cucumber.io/docs/gherkin/languages/#gherkin-dialect-nl-content still mentions "Rule" instead of "Regel" for the nl language. 

The docs there aren't directly tied to the active codebase. See https://github.com/cucumber/docs/issues/936 raised recently to tackle this.

> Wouldn't it be better if both "Rule" and "Regel" are possible in the nl language feature? @Coedice @luke-hill

This goes against the language paradigms of gherkin. If you want to mix-n-match your language prefs then gherkin won't be able to tell them apart. You can use `*` as a language agnostic keyword to tackle this if you'd prefer?

smuller-nl:
Thank you @luke-hill for your swift reply. We had lost some time hunting for the reason why our feature files didn't parse anymore, and the change log and the localisation docs is where we had looked, but didn't find anything. My suggestions were just to prevent others losing time if in the future similar changes are made. Implementing cucumber/docs#936 would certainly help!

PR #176, opened 2023-09-24
[Perl] Generated Languages.pm as part of the build process

### LINKED ISSUE #32: perl: Languages.pm is not mergable

Currently the generated `Languages.pm` consists of a single line.

https://github.com/cucumber/gherkin/blob/1cc2ef0153603c5e51818b59ff6d632238491226/perl/lib/Gherkin/Generated/Languages.pm#L1-L7

As a result, when multiple contributors submit changes to `gherkin-languages.json` their PR's will always conflict. Because PRs are not often looked at this happens more often then you'd think.

Consider changing `build_languages.pl` so it generates the file with some formatting.

ehuelsmann:
Fixes #32

### 🤔 What's changed?

The Languages.pm file is now generated upon building the release artifacts or when running the test suite.

Fixes: #32 (Languages.pm not mergeable)

Does this address the concern sufficiently?

### 📋 Checklist:

----


perl/dist.ini
```diff
@@ -36,10 +36,16 @@ exclude_filename=gherkin-perl.razor
...
 filename=CHANGELOG.md
 
+[GatherFile / GeneratedLanguages]
+filename=lib/Gherkin/Generated/Languages.pm
+
 [Hook::VersionProvider]
 . = my $v = `cat ./VERSION`; chomp( $v ); $v;
+
+[Run::BeforeBuild]
+run = helper-scripts/build_languages.pl >lib/Gherkin/Generated/Languages.pm
```
mpkorstanje, line 51: With this in place, I would expect the code from the Makefile that does the generation is no longer needed. And with it some other changes too.

```Makefile
lib/Gherkin/Generated/Languages.pm:
	perl helper-scripts/build_languages.pl > $@
```

ehuelsmann: Currently the Makefile targets are still used for the GitHub Actions CI runs. Those runs are specialized because they depend on the conformance data, which isn't packaged in the release tarball. I need to check again, but from what I remember `dzil` passes a cleaned environment to the test scripts, so the scripts needed a predictable directory to find the conformance data.

The `dzil build` target only packages the files for release. The `dzil test` target only runs a few syntax checks. Definitely not enough to validate conformance.

mpkorstanje: Yes, the Makefile as a whole is still used. 

But you've created an alternative for the generation of `lib/Gherkin/Generated/Languages.pm`. And I would generally expect a build process to have a single way of doing this. So I would not expect `perl helper-scripts/build_languages.pl ` to be invoked in the Makefile anymore.

Perhaps `dzil build` would have to be invoked instead `cpanm`? And perhaps the acceptance tests would have to be ran with the artifact produced by `dzil build`? 

```
perl5:
	cpanm --notest --local-lib ./perl5 --installdeps --with-develop .
```

ehuelsmann: Running the tests with an environment as it would have been created by installing the artifact from `dzil build` , can be done using `dzil test`  (to run the Perl test suite) or `dzil run` (to run an arbitrary command).

I guess what it comes down to is: should the Makefile follow the patterns all other Makefiles use (thereby deviating from "the Perl way")? Or should I use "the Perl way", thereby deviating from "the project way"?

I guess I know the answer: since Elixir doesn't do it the project way anyway, so there's precedent for doing it "the language's ecosystem way". I'll change this PR to de-duplicate accordingly.

mpkorstanje: I'm not sure I know what the "Perl way" is. ;)

But most languages do follow the pattern of executing the code in the artifact produced by the build system - occasionally with a shim.

For example JavaScript uses npm to build the artifact.

```Makefile
.built: node_modules dist $(SOURCE_FILES)
	touch $@

node_modules:
	npm install

dist:
	npm run build
```

And then to run the tests 

```
GHERKIN = node ./test-cli.mjs
```
Where `test-cli.mjs` is a shim that uses Gherkin to run the acceptance tests.

Perl seems to differ mostly in that it uses both `cpanm` (to create the artifact used by tests) and `dzil` (to create the released artifact). I would prefer it if we have system for both.

ehuelsmann: > Perl seems to differ mostly in that it uses both cpanm (to create the artifact used by tests)

Ah! There'sa misconception there. Good to be able to clarify that. `cpanm` is actually not building the artifact (`dzil build` is); `cpanm` is about installing the dependencies to be able to run the tests (as well as `dzil` itself). Until this PR (which removes `lib/Gherkin/Generated/Languages.pm`) the tests can be run without any build steps straight out of the repository.

This PR changes that, because with this PR a build step *is* required (generating the missing languages file) before the tests can be run. This is an extra step being introduced in this PR.

My plan now is to change the `acceptance` Makefile target: it will invoke `dzil test`, which will generate the missing languages file. I will also change the `acceptance/*` targets to Perl code which can be run as part of `dzil test` (which assumes it can run the files in t/ as Perl test files; it doesn't invoke the project's Makefile).

mpkorstanje: > I will also change the acceptance/* targets to Perl code which can be run as part of dzil test

This part has me a little confused. The way the acceptance tests are run seems correct to me.

>  Until this PR (which removes lib/Gherkin/Generated/Languages.pm) the tests can be run without any build steps straight out of the repository.

Then I would now expect `perl helper-scripts/build_languages.pl` to be replaced with `dzil build` or something like that.

ehuelsmann: > This part has me a little confused. The way the acceptance tests are run seems correct to me.

They are as long as the file `lib/Gherkin/Generated/Languages.pm` exists, but this commit removes that file. Instead, I'm invoking `dzil build`, which generates a release tarball, but Perl can't use that tarball to load its modules from. It also generates a directory with the content of that tarball, but that has the version number embedded (making the Makefile a more complex).

To keep the makefile simple, I was going to go with the way Elixir works, but that breaks down for other reasons (the URI value).

I've taken yet another direction with this PR now, based on your comments and my findings so far.

ehuelsmann:
The latest Iteration of this PR generates Languages.pm before testing and release artefact building. It does that by extending the `dzil build` command. The tests are being run based on what is in the git tree (as they always have). I

f we want to run the tests based on what is in the release tarball, we need to install its content. `dzil test` and `dzil run` exists for that purpose: they create an out of tree isolated installation as would be achieved with a vanilla installation. Using dzil this way is not possible, because our testdata needs the tests to run in a relative path to the test data set.

Looking at how Ruby does it, I see it uses `bundler `. Although I'm not too much into the details, I think the description of what we do now (run from the working tree) is exactly what Ruby does too, since `bundle package` isn't required for `bundle exec` and the latter is what the tests use.

Concluding: I think this PR brings the Perl situation in line with what Ruby does.

ehuelsmann:
Please let me know what I can do to expedite this PR, or let me know what other direction to take. The PR is created in response to the feedback on the Perl approach with a pre-built Languages.pm file in the repository (which is neither what other languages do, nor desirable, as per the feedback). If a different approach is required, I'll work on that.

mpkorstanje:
Cheers! I've got next week planned for core work. So, pending availability, hopefully by then.

mpkorstanje (changes requested):
Hey Eric thanks for keeping up with this! It has taken some time. :sweat_smile: 

I've cleaned up the `Makefile` and pushed some changes so we can run the acceptance tests against the artifact produces by `dzil build`. 

One item I'm not sure how to fix yet is running the "unit test" before the acceptance tests. I think it should work with `dzil test` but I'm not sure how to get that working in a GitHub action. 

Could you help with that?

mpkorstanje (approved):
Okay, got something that isn't horrible to work. 

Let me know if you see I've committed any sins against Perl. :wink:

ehuelsmann:
@mpkorstanje looking very nice! Thanks for ironing out the Makefile stuff and shaping a bit more to your linking. No sins againt Perl paradigms that I can see. I'd say: if it doesn't add new failures, this is an improvement. Let's merge it!

ehuelsmann:
Thanks for merging! Another action item off my list.

PR #228, opened 2024-02-29
CI: Update latest PHP version to last non-EOL version

olleolleolle:
### 🤔 What's changed?

Went from 8.1 to 8.2.

### ⚡️ What's your motivation? 

8.1 is EOL.
8.2 is still supported.

When reading https://phpunit.de/announcements/phpunit-11.html I saw that since Feb 2 2024, 8.2 is the oldest still-supported PHP version.

### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

### ♻️ Anything particular you want feedback on?

-

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://cucumber.io/conduct/)

PR #229, opened 2024-02-29
Update in composer php

olleolleolle:
This PR is a follow-up to #228.

### 🤔 What's changed?

Make the dependency on a non-EOL'd PHP version manifest in composer.json.

A good comment in #228.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 
 ## [Unreleased]
 
+### Changed
+- [PHP] Require PHP 8.2+ in CI and in composer.json.
+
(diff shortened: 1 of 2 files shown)


mpkorstanje:
@ciaranmcnulty do you think we have a prayer at upgrading this?

olleolleolle:
@mpkorstanje To me, it seemed like psalm isn't yet ready, but perhaps we can make it ready, by creating a psalm that supports the new parser version?

mpkorstanje:
Thanks for looking into this. It looks like [PHP 8.1 will still get updates](https://www.php.net/supported-versions.php) so I think we can wait for Psalm to upgrade their parser.

mpkorstanje:
It starts to work again on PHP 8.3 but does come with a lot of new problems found by Psalm.

PR #237, opened 2024-03-14
Build every language on changes to the GitHub workflows

mpkorstanje:
### 🤔 What's changed?

Build every language on changes to the GitHub workflows. 

We usually only run CI for the languages that changed because it is quite slow. But @olleolleolle was quite confused when CI didn't run for his changes (https://github.com/cucumber/gherkin/pull/233#issuecomment-1991341348). This should prevent that from happening.


olleolleolle:
Thanks for working on reducing my confusion!

PR #238, opened 2024-03-15
[php] Try to follow Psalm tips

olleolleolle:
### 🤔 What's changed?

This PR is an experiment run, to see if Psalm's "find unused" can be tamed, to become useful.

Psalm seems like a cool tool.

Is this a viable path?

I downgraded a Property in a constructor to a regular positional parameter – this was due to a Psalm warning. Is that reasonable?

One error handler dealt with return values, when it looked like that wasn't what it usually did. I turned it into a regular "callbacky" thing, instead.

Well, any thoughts around keeping these Psalm checks enabled would be welcome. Also thoughts around "why this does not suit Gherkin".

### 📋 Checklist:

### DIFF
--- a/php/src/AstNode.php
+++ b/php/src/AstNode.php
@@ -31,11 +31,11 @@ final class AstNode
     /**
      * @template T of object
      *
-     * @param class-string<T> $expectedType
+     * @param class-string<T> $_expectedType
      *
      * @psalm-return list<T>
      */
-    public function getItems(string $expectedType, RuleType $ruleType): array
+    public function getItems(string $_expectedType, RuleType $ruleType): array
     {
         $items = $this->subItems[$ruleType->name] ?? [];
 
--- a/php/src/AstNode.php
+++ b/php/src/AstNode.php
@@ -31,12 +31,13 @@ final class AstNode
     /**
      * @template T of object
      *
-     * @param class-string<T> $_expectedType
+     * @param class-string<T> $expectedType
      *
      * @psalm-return list<T>
      */
-    public function getItems(string $_expectedType, RuleType $ruleType): array
+    public function getItems(string $expectedType, RuleType $ruleType): array
     {
+        $expectedType == 0; // Avoid error: Param #1 is never referenced in this method (see https://psalm.dev/135)
         $items = $this->subItems[$ruleType->name] ?? [];
 
         /**
--- a/php/src/GherkinParser.php
+++ b/php/src/GherkinParser.php
@@ -18,6 +18,8 @@ use Generator;
 
 /**
  * Parses a Gherkin document (or list of Source envelopes) and emits Cucumber Messages envelopes
+ *
+ * @psalm-api
  */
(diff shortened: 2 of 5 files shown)


php/src/AstNode.php
```diff
@@ -31,11 +31,11 @@ public function add(RuleType $ruleType, object|array|string $entry): void
     /**
      * @template T of object
      *
-     * @param class-string<T> $expectedType
+     * @param class-string<T> $_expectedType
      *
      * @psalm-return list<T>
      */
-    public function getItems(string $expectedType, RuleType $ruleType): array
+    public function getItems(string $_expectedType, RuleType $ruleType): array
```
mpkorstanje, line 38: I'm not familiar with php, what does the `_` do?

php/src/GherkinParser.php
```diff
@@ -28,13 +30,13 @@ final class GherkinParser
      * @param bool $predictableIds Ignored if IdGenerator is provided
      */
     public function __construct(
-        private readonly bool $predictableIds = false,
+        bool $predictableIds = false,
```
mpkorstanje, line 33: This looks okay to me.

php/src/Parser/ParserTrait.php
```diff
@@ -72,12 +72,10 @@ private function addError(ParserContext $context, ParserException $error): void
...
      * @template U
      * @param callable() : U $action
-     *
-     * @return U
      */
-    private function handleAstError(ParserContext $context, callable $action): mixed
+    private function handleAstError(ParserContext $context, callable $action): void
     {
-        return $this->handleExternalError($context, $action, null);
+        $this->handleExternalError($context, $action, null);
```
mpkorstanje, line 78: This looks okay to me too.

php/src/AstNode.php
olleolleolle: It is a naming convention to mean "I'm not using this positional argument".

mpkorstanje: Does php allow positional arguments to be named? I.e. would this break someone using this as a named argument?

olleolleolle: @mpkorstanje Oh I believed things about PHP which were true before 8.0.

https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments

> PHP 8.0.0 introduced named arguments as an extension of the existing positional parameters. Named arguments allow passing arguments to a function based on the parameter name, rather than the parameter position. This makes the meaning of the argument self-documenting, makes the arguments order-independent and allows skipping default values arbitrarily.

This makes my change wrong!

olleolleolle: Well, I followed the change tip in the Psalm note:

https://psalm.dev/docs/running_psalm/issues/UnusedParam/

> Can be suppressed by prefixing the parameter name with an underscore:

olleolleolle: Referring to the parameter, that works: 
411ea6d

PR #243, opened 2024-04-10
Provide trailing space in Irish keywords

kieran-ryan:
### 🤔 What's changed?

- Spaces are used at the end of step keywords e.g. `Given ` and have been applied to the Irish translation
- Applied translations that were partially applied across language implementations

- As spaces were missing, and with the language server using the official gherkin parser; steps were incorrectly marked undefined when a step was introduced between the step keyword and the step text; as the space was being considered part of the step text, instead of typically part of the step keyword. This resolves that issue
- Ensure language translations are fully aligned and up to date

Screenshot 2024-04-10 at 22 11 33

### 📋 Checklist:

----

### DIFF
--- a/gherkin-languages.json
+++ b/gherkin-languages.json
@@ -278,7 +278,7 @@
       "Nə vaxt ki "
     ]
   },
-    "be": {
+  "be": {
     "and": [
       "* ",
       "I ",
@@ -1401,14 +1401,14 @@
   "ga": {
     "and": [
       "* ",
-      "Agus"
+      "Agus "
     ],
     "background": [
       "Cúlra"
     ],
     "but": [
       "* ",
-      "Ach"
+      "Ach "
     ],
     "examples": [
       "Samplaí"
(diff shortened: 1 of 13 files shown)


gherkin-languages.json
```diff
@@ -3134,15 +3134,15 @@
...
+      "மேலும் ",
       "மற்றும் "
     ],
     "background": [
       "பின்னணி"
     ],
     "but": [
       "* ",
-      "ஆனால்  "
+      "ஆனால் "
```
mpkorstanje, line 3145: How did you spot these?

mpkorstanje (changes requested):
Looks good, also looks like you fixed something in Tamil. That would be worth it's own changelog note.

gherkin-languages.json
kieran-ryan: Tried the Irish implementation, observed the issue, and then ran a check against all keywords where we typically expect a trailing space.

There were other occurrences (such as for emojis and CJK characters) though I wouldn't know the expectation in these cases; particularly with [valid use cases such as French](https://github.com/cucumber/gherkin/blob/f5cbb5d7f52de872eb881526ea5eff4a980bbd7e/gherkin-languages.json#L1361) where we'd avoid a trailing space.

We could potentially add to each language test job (or standalone job) to check a deep match against the source of truth language configuration. Unsure of the tradeoffs, but would be a nice peace of mind to automate alignment.

Good shout - I'll introduce an additional changelog entry - thanks!

PR #272, opened 2024-08-16
[.NET] Adopt c# 12 primary constructors

Romfos:
### 🤔 What's changed?

- Adopt c# 12 primary constructors
documentation: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors
- Remove some unsued private settters

### ⚡️ What's your motivation? 

Adopt modern c# language feature to simplify code base

### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)

### ♻️ Anything particular you want feedback on?

no changes in product. code style changes only

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)
- [ ] I've changed the behaviour of the code
  - [ ] I have added/updated tests to cover my changes.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
- [ ] Users should know about my change
  - [x] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

PR #275, opened 2024-08-27
[.NET] Adopt some modern c# language feature, enable warnings as errors, adopt system.text.json

Romfos:
### 🤔 What's changed?

- Enable warnings as errors
- Replace TinyJson to System.Text.Json
- Adopt c# 10 ImplicitUsings
- Simplify supported target frameworks: net8.0;netstandard2.0 (.NET Framework 4.6.2 will be supported via .NET Standard)
- Add tests for .NET Framework

Original goal of this PR was adopt modern c# and enable warnings as errors, but I've found that some service inside this JSONParser is obsolete:

https://github.com/cucumber/gherkin/pull/272/files
Check warning on line 335 in dotnet/Gherkin/TinyJson/JSONParser.cs
'FormatterServices' is obsolete: 'Formatter-based serialization is obsolete and should not be used.' (https://aka.ms/dotnet-warnings/SYSLIB0050)

Because modern .NET already has built-in support for system.text.json make sense to use it by default and don't reinvent  the wheel and tring to fix and support in future this custom parser.
for .NET Framework system.text.json is provided via nuget package

n\a 

### 📋 Checklist:

----


Romfos:
discussion point:
I've read this discussion in https://github.com/cucumber/messages/pull/237#issuecomment-2225810555
by this reason I've decided to use version 6.0.0 for system.text.json

context:
In second half of 2024 most of developers are using .NET 6 as primary target platform and step by step migrating to .NET 8 (.NET 6 EOL in December 2024)

my private opinion:
- Make sense to have .NET6/8/9 as primary target platforms and support more features on these platforms
- .NET Framework is still make sense to support, but as second priority
- In case of conflict (as in case with System.Text.Json) make sense to choose more optimal solution for modern .NET rather then .NET Framework

by this reasons I would prefer to see latest System.Text.Json with version 8, even if for .NET Framework it will be maybe less optimal

dotnet/Gherkin/Gherkin.csproj
```diff
@@ -31,8 +32,12 @@
     <PackageOutputPath>bin/$(Configuration)/NuGet</PackageOutputPath>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Text.Json" Version="6.0.0" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
mpkorstanje: Nevermind. Found the usage.

dotnet/Gherkin/GherkinDialectProvider.cs
```diff
@@ -55,8 +52,7 @@ protected virtual Dictionary<string, GherkinLanguageSetting> LoadLanguageSetting
 
     protected virtual Dictionary<string, GherkinLanguageSetting> ParseJsonContent(string languagesFileContent)
     {
-        // ReSharper disable once InvokeAsExtensionMethod
-        return JSONParser.FromJson<Dictionary<string, GherkinLanguageSetting>>(languagesFileContent);
+        return JsonSerializer.Deserialize<Dictionary<string, GherkinLanguageSetting>>(languagesFileContent, new JsonSerializerOptions(JsonSerializerDefaults.Web));
```
mpkorstanje, line 55: For the Java implementation I removed the dependency on serialization by generating the source code for the Gherkin dialects file. You may want to consider a similar approach. Not necessarily in this pull request.

- https://github.com/cucumber/gherkin/blob/main/java/src/main/groovy/io/cucumber/gherkin/GherkinDialects.groovy
- https://github.com/cucumber/gherkin/blob/main/java/src/main/groovy/io/cucumber/gherkin/GherkinDialects.gsp

mpkorstanje: That is reasonable, but then https://github.com/cucumber/messages/pull/237#issuecomment-2225649432 will have to be addressed.

Romfos: ok, I've added test for that

mpkorstanje: Effectively yes.

The upgrade will fail, so Renovate will raise a PR instead. Then we close the PR and Renovate will take note and not offer the same upgrade again.

mpkorstanje (approved):

dotnet/Gherkin/GherkinDialectProvider.cs
mpkorstanje: And you can see it in action here https://github.com/cucumber/gherkin/pull/279

Romfos: Hm, it will update all nuget packages very time or 1 PR = 1 nuget package?

It could be a problem
https://github.com/cucumber/gherkin/pull/278

mpkorstanje: It should be one PR for each dependency unless there is a a profile that bundles them. These profiles are manually written. So usually quite sane.

PR #276, opened 2024-08-29
[.NET] - AstMessagesConverter to return a default Location of (0,0) 

clrudolphi:
when the AST does not include a location for an element

I

n Reqnroll and SpecFlow, plugins may provide dynamically created Example data rows that don't have source locations. Hard-coding these to (0, 0).

### 📋 Checklist:

- [ X] I've changed the behaviour of the code
  - [X ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

### DIFF
--- a/dotnet/Gherkin/CucumberMessages/AstMessagesConverter.cs
+++ b/dotnet/Gherkin/CucumberMessages/AstMessagesConverter.cs
@@ -74,6 +74,9 @@ public class AstMessagesConverter
 
     private static Location ConvertLocation(Ast.Location location)
     {
+        // In Reqnroll and SpecFlow, plugins may provide dynamically created Example data rows that don't have source locations. Hard-coding these to (0, 0).
+        if (location == null)
+            return new Location(0, 0);
         return new Location(location.Column, location.Line);
     }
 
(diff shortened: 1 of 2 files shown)


dotnet/Gherkin/CucumberMessages/AstMessagesConverter.cs
```diff
@@ -74,6 +74,9 @@ private Feature ConvertFeature(Ast.GherkinDocument gherkinDocument)
 
     private static Location ConvertLocation(Ast.Location location)
     {
+        // In Reqnroll and SpecFlow, plugins may provide dynamically created Example data rows that don't have source locations. Hard-coding these to (0, 0).
+        if (location == null)
```
mpkorstanje, line 78: This sounds like a leaky abstraction. Shouldn't the callers of this method perform this operation? Gherkin itself does not have a concept it dynamically created example rows.

clrudolphi:
Withdrawing this pull request. I've found another solution directly within Reqnroll to address this.

mpkorstanje:
No problem.

PR #281, opened 2024-09-11
[.NET] Add initial nullable reference types support

Romfos:
### 🤔 What's changed?

- Enable nullable reference types

- Support more .NET feature on modern runtimes https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references

n\a

### 📋 Checklist:

  - [X] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

### DIFF
--- a/dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
+++ b/dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
@@ -174,21 +174,18 @@ public class PickleCompiler(IIdGenerator idGenerator)
     protected virtual StepKeywordType GetKeywordType(Step step, StepKeywordType lastStepKeywordType)
     {
         var stepKeywordType = step.KeywordType;
-        switch (stepKeywordType)
+
+        return step.KeywordType switch
         {
-            case StepKeywordType.Context:
-            case StepKeywordType.Action:
-            case StepKeywordType.Outcome:
-            case StepKeywordType.Unknown:
-                return stepKeywordType;
-            case StepKeywordType.Conjunction:
-                return lastStepKeywordType == StepKeywordType.Unspecified ||
-                       lastStepKeywordType == StepKeywordType.Unknown
-                    ? StepKeywordType.Context
-                    : lastStepKeywordType;
-            default:
-                return StepKeywordType.Unspecified;
-        }
+            StepKeywordType.Context => StepKeywordType.Context,
+            StepKeywordType.Action => StepKeywordType.Action,
+            StepKeywordType.Outcome => StepKeywordType.Outcome,
+            StepKeywordType.Unknown => StepKeywordType.Unknown,
+            StepKeywordType.Conjunction => lastStepKeywordType is StepKeywordType.Unspecified or StepKeywordType.Unknown
+                ? StepKeywordType.Context
+                : lastStepKeywordType,
+            _ => StepKeywordType.Unspecified
+        };
     }
 
 
--- a/dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
+++ b/dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
@@ -165,7 +165,7 @@ public class PickleCompiler(IIdGenerator idGenerator)
         }
     }
 
-    protected virtual PickleStep CreatePickleStep(Step step, string text, PickleStepArgument argument, IEnumerable<string> astNodeIds)
+    protected virtual PickleStep CreatePickleStep(Step step, string text, PickleStepArgument? argument, IEnumerable<string> astNodeIds)
     {
         CurrentStepKeywordType = GetKeywordType(step, CurrentStepKeywordType);
         return new PickleStep(argument, astNodeIds, idGenerator.GetNewId(), text, CurrentStepKeywordType);
@@ -192,13 +192,13 @@ public class PickleCompiler(IIdGenerator idGenerator)
     }
 
 
-    protected virtual PickleStepArgument CreatePickleArgument(Step argument)
+    protected virtual PickleStepArgument? CreatePickleArgument(Step argument)
     {
         var noCells = Enumerable.Empty<TableCell>();
         return CreatePickleArgument(argument, noCells, noCells);
     }
 
-    protected virtual PickleStepArgument CreatePickleArgument(Step step, IEnumerable<TableCell> variableCells, IEnumerable<TableCell> valueCells)
+    protected virtual PickleStepArgument? CreatePickleArgument(Step step, IEnumerable<TableCell> variableCells, IEnumerable<TableCell> valueCells)
     {
         if (step.DataTable != null)
         {
--- a/dotnet/Gherkin/CucumberMessages/Types/PickleStep.cs
+++ b/dotnet/Gherkin/CucumberMessages/Types/PickleStep.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Serialization;
 
 namespace Gherkin.CucumberMessages.Types;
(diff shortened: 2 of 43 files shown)


Romfos:
#nullable is still disabled for:
AstBuilder
Token

reasons: 
1) It required to do changes in Parser that is code generated and will require update BERP
2) AstBuilder.cs and Token.cs have potential a lot of "null pointer exception" problems that are hidden now. Fix will lead to changes that could be not so backward compatible

Romfos:
Important:
This PR is a "source breaking change". 
Nullable annotations could lead to new warnings in code that depends on this library if external code will try to pass in not expected places

dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
```diff
@@ -192,13 +192,13 @@ protected virtual StepKeywordType GetKeywordType(Step step, StepKeywordType last
     }
 
 
-    protected virtual PickleStepArgument CreatePickleArgument(Step argument)
+    protected virtual PickleStepArgument? CreatePickleArgument(Step argument)
```
clrudolphi, line 195: Why would the return type here be nullable? The method will definitely return an instance of PickleStepArgument.

dotnet/Gherkin/CucumberMessages/Types/PickleStep.cs
```diff
@@ -1,28 +1,31 @@
...
-    public string Id { get; set; }
+    public required string Id { get; set; }
 
     [DataMember(Name = "text")]
-    public string Text { get; set; }
+    public required string Text { get; set; }
 
     [DataMember(Name = "type")]
-    public StepKeywordType Type { get; set; }
+    public required StepKeywordType Type { get; set; }
```
clrudolphi, line 21: According to the schema for Cucumber Messages, the Type property of a PickleStep is not required.

dotnet/Gherkin/CucumberMessages/Types/SourceReference.cs
```diff
@@ -5,10 +5,10 @@ namespace Gherkin.CucumberMessages.Types;
 public class SourceReference
 {
```
clrudolphi, line 6: A SourceReference should contain one of a Location, Uri, JavaMethod or a JavaStackTraceElement.

dotnet/Gherkin/CucumberMessages/Types/Step.cs
```diff
@@ -5,23 +5,23 @@ namespace Gherkin.CucumberMessages.Types;
 public class Step
 {
```
clrudolphi, line 6: According to the schema, the only required elements of a Step are: Location, Keyword, Text, and Id.

clrudolphi (changes requested):
I've reviewed the DTOs created by the AstConverter against the schema for Messages and made a few comments where I found differences.
The AstConverter is a limited implementation of Cucumber.Messages. Eventually we should port this over consuming the Cucumber.Messages types directly.
@Romfos - would love to have your assistance in applying this same nullable annotations to the template that is used to generate the Io.Cucumber.Messages.Types classes.

dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
Romfos: because on line 240 this method return null

dotnet/Gherkin/CucumberMessages/Types/SourceReference.cs
Romfos: thank you for feedback. I will make an optimal it then

dotnet/Gherkin/CucumberMessages/Types/Step.cs
Romfos: what about KeywordType? should it be optional? as I see - this is only 1 difference

dotnet/Gherkin/CucumberMessages/Types/PickleStep.cs
Romfos: thank you for feedback, fixed

Romfos:
could you please review all code base, not only changed files

motivation:
```
public abstract class StepsContainer(Location location, string keyword, string name, string description, Step[] steps)
    : IHasLocation, IHasDescription, IHasSteps
{
    public Location Location { get; } = location;
    public string Keyword { get; } = keyword;
    public string Name { get; } = name;
    public string Description { get; } = description;
    public IEnumerable<Step> Steps { get; } = steps;
}
```

should description be nullable here?

Romfos:
> e your assistance in applying this same nullable annotations to the template that is used to generate the Io.Cucumber.Messages.Types classes.

hm, we have a separate library with copy-paste here?

dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
clrudolphi: No, this method definitely returns a PickleStepArgument. The overload that starts on line 201 returns a possible null value, but not this one. I'm not sure how much headache is caused by such a difference.

dotnet/Gherkin/CucumberMessages/Types/Step.cs
clrudolphi: According to the schema, KeywordType, Docstring, and DataTable are optional.

clrudolphi:
> please, during review, pay attention for all code base, not only changed files
> 
> motivation:
> 
> ```
> public abstract class StepsContainer(Location location, string keyword, string name, string description, Step[] steps)
>     : IHasLocation, IHasDescription, IHasSteps
> {
>     public Location Location { get; } = location;
>     public string Keyword { get; } = keyword;
>     public string Name { get; } = name;
>     public string Description { get; } = description;
>     public IEnumerable<Step> Steps { get; } = steps;
> }
> ```
> 
> should description be nullable here?

I'm really not sure. This part of the code base concerns the AST; I'm not knowledgable enough about that section to provide guidance. Gaspar is the expert to depend upon.

clrudolphi:
> > e your assistance in applying this same nullable annotations to the template that is used to generate the Io.Cucumber.Messages.Types classes.
> 
> hm, we have a separate library with copy-paste here?

I suspect its a result of how and when the libraries were built. They are not copy-paste, but parallel implementations of the same object model. The one you see in this library is a subset of that in the Cucumber.Messages library.

dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
clrudolphi: Ah, yes, I had overlooked that. Withdrawn.

dotnet/Gherkin/CucumberMessages/Types/Step.cs
Romfos: fixed

Romfos:
> I've reviewed the DTOs created by the AstConverter against the schema for Messages and made a few comments where I found differences. The AstConverter is a limited implementation of Cucumber.Messages. Eventually we should port this over consuming the Cucumber.Messages types directly. @Romfos - would love to have your assistance in applying this same nullable annotations to the template that is used to generate the Io.Cucumber.Messages.Types classes.

https://github.com/cucumber/messages/pull/253

Romfos:
Probably we need to reference them from Cucumber.Messages insted of have this non-compatible copy here

mpkorstanje:
The latest version of messages has been released. Should be available on NuGet.

https://github.com/cucumber/messages/releases/tag/v26.0.1

Romfos:
ok, then lets put on hold, until cucumber messages are not updated

PR #286, opened 2024-09-21
Apply `black` code formatter to the python codebase

youtux:
### 🤔 What's changed?
* Apply `black` code formatter to the python codebase, to enforce a standard formatting
* Add a `.pre-commit-config.yaml` to the project, so that we can enforce this convention
* TODOs
  - [ ] ~~add github action to check the pre-commit hooks have been applied~~ Will add [pre-commit.ci](https://pre-commit.ci) app once this lands
  - [x] merge https://github.com/cucumber/gherkin/pull/283 first

The python codebase does not really use a standard formatting, this is arguably widely accepted as "annoying".

### 📋 Checklist:

----


mpkorstanje (commented):
I see that you are adding pre-commit hooks. Are these setup in such a way that it will not interfere with contributors who make contributions without python installed?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 24.8.0
+    hooks:
+      - id: black
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        language_version: python3.8
+        exclude: "python/gherkin/parser.py"
```
mpkorstanje, line 14: Ideally this is kept in the `python` folder.

youtux: done in f1e0e7e2

youtux:
> I see that you are adding pre-commit hooks. Are these setup in such a way that they will not interfere with contributors who make contributions without python installed?

yes

jsa34 (approved):
Looks good to me

jsa34:
I'm assuming the GitHub actions work will be in separate PR

.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
```
mpkorstanje, line 3: Would it be possible to keep this whole file in the `python` folder?

mpkorstanje (changes requested):
I'm not sure about the pre-commit hook yet. Looks like it will activate regardless.

youtux:
> One thing to bare in mind (I'm doing the same in ruby), is some long methods in gherkin are written in a way that is almost language agnostic, so it reads the same in every language.

not sure how this applies in this PR @luke-hill

.pre-commit-config.yaml
youtux: pre-commit installs hooks in a completely isolated environment that does not affect global packages. If other contributors don't touch the `python` folder, the `black` hook won't even run at all.

youtux: I tried putting `.pre-commit-config.yaml` under the `python` directory, but it won't allow any other `.pre-commit-config.yaml` to be used if anyone decides to have one under `java/` or any other folder.
Seems like a limitation of pre-commit.
So the only way is to keep it on the root dir, and explicitly whitelist the dirs that each hook should handle.

PR #300, opened 2024-10-15
Add support for .NET Framework 4.7.2

DoubleCouponDay:
### 🤔 What's changed?

This Pull request adds support for legacy .NET Framework 4.7.2 apps such as Unity, Godot 3 and more. 

It is a simple fix by using the `Newtonsoft.Json` nuget package for serializing and deserializing Json strings.

I love using Gherkin for Unit Testing, but it doesn't work in my Godot 3 project.

### 📋 Checklist:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 - [PHP, Java, Ruby, JavaScript] update dependency messages up to v26
 - [Python] Added type annotations ([#283](https://github.com/cucumber/gherkin/pull/283))
 - [Python] Switch to pyproject.toml ([#290](https://github.com/cucumber/gherkin/pull/290))
+- [.NET] Add Support for .NET Framework 4.7.2 [#300](https://github.com/cucumber/gherkin/pull/300)
 
 ### Changed
 - [.NET] Drop unsupported frameworks. Now supported target frameworks are .NET 8, .NET Standard 2.0 ([#265](https://github.com/cucumber/gherkin/pull/265))
--- a/dotnet/Gherkin.Specs/CLI/Program.cs
+++ b/dotnet/Gherkin.Specs/CLI/Program.cs
@@ -1,8 +1,7 @@
 using Gherkin.CucumberMessages;
 using Gherkin.Specs.EventStubs;
 using Gherkin.Specs.Tokens;
-using System.Text.Json;
-using System.Text.Json.Serialization;
+using Newtonsoft.Json;
 
 namespace Gherkin.Specs.CLI;
 
@@ -100,11 +99,7 @@ class Program
         {
             foreach (var evt in gherkinEventsProvider.GetEvents(sourceEventEvent))
             {
-                var jsonString = JsonSerializer.Serialize(evt, new JsonSerializerOptions(JsonSerializerDefaults.Web)
-                {
-                    Converters = { new JsonStringEnumConverter() },
-                    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
-                });
+                var jsonString = JsonConvert.SerializeObject(evt);
                 Console.WriteLine(jsonString);
             }
         }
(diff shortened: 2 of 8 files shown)


mpkorstanje:
Looks like the acceptance tests aren't passing. It seems the serialization isn't configured correctly.

Before fixing that it would be useful to get a consensus on what versions of .Net to support though.  In https://github.com/cucumber/gherkin/pull/275 it is noted that .NET Framework 4.6.2 will be supported via .NET Standard. So I would expect 4.7.2 to be covered that way as well.

Note: I have no in debt .Net knowledge. I'll defer to @gasparnagy for that.

Romfos:
Hi,

What is motivation for that?

Why did you replace System.Text.Json to Newtonsoft.Json?
Now we support NETFX and run tests for for .NET FX 4.6.2 and we don't need a separate tfm for 4.7.2

Something didn't work for you? Maybe you can put some details and we can help you without merging this?

Thank you

gasparnagy:
I can just agree to @Romfos, I don't see why we would need this. .NET Framework v4.6.2+ (including .NET Framework v4.7.2) does support .NET Standard 2.0 that we target, so we do not need a separate target for it.

We use the library in Reqnroll for .NET Framework v4.7.2 without any issues.

So I think this must be some misunderstanding.

DoubleCouponDay:
Hello and thank you for your interest in my pull request. My motivation for making these changes was to satisfy a runtime error in a unit test that uses `Gherkin=5.0.0` and `Xunit.Gherkin.Quick=29.0.0`. When I run the test explorer using Visual Studio, I get the following error message. Since the Gherkin assembly could not be found, I assumed that a legacy project, such as .NET Framework 4.7.2, was not compatible.

When I cloned the Gherkin project and set the target runtimes as `net472;net8.0;net462`, the project failed to build. I was unaware that there was a `System.Text.Json` nuget package available, if that is your preference.

```
System.InvalidOperationException : Exception during discovery:
System.IO.FileNotFoundException: Could not load file or assembly 'Gherkin, Version=5.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.
File name: 'Gherkin, Version=5.0.0.0, Culture=neutral, PublicKeyToken=null'
   at Xunit.Gherkin.Quick.ScenarioDiscoverer.<Discover>d__2.MoveNext()
   at Xunit.Sdk.XunitTestFrameworkDiscoverer.FindTestsForMethod(ITestMethod testMethod, Boolean includeSourceInformation, IMessageBus messageBus, ITestFrameworkDiscoveryOptions discoveryOptions) in /_/src/xunit.execution/Sdk/Frameworks/XunitTestFrameworkDiscoverer.cs:line 119
   at Xunit.Sdk.XunitTestFrameworkDiscoverer.FindTestsForType(ITestClass testClass, Boolean includeSourceInformation, IMessageBus messageBus, ITestFrameworkDiscoveryOptions discoveryOptions) in /_/src/xunit.execution/Sdk/Frameworks/XunitTestFrameworkDiscoverer.cs:line 135
```

image

DoubleCouponDay:
Using `System.Text.Json` as a namespace reference, I received a compilation error for the `net472` target. I have yet to see how this library can support that.

image

Romfos:
According to error for me it looks like problem with binding redirects when 1 library target version 5 and then you install separately version 29

Error doesn't say that lib don't support .NET Framework

My proposition:
1) Uninstall there 2 libraries
2) Try to install them again via nuget explorer in visual studio and check binding redirects

binding redirects should use higher version

DoubleCouponDay:
It seems I have not regenerated the parsers. I will attempt to get that working tonight.

DoubleCouponDay:
> According to error for me it looks like problem with binding redirects when 1 library target version 5 and then you install separately version 29
> 
> Error doesn't say that lib don't support .NET Framework
> 
> My proposition:
> 
> 1. Uninstall there 2 libraries
> 2. Try to install them again via nuget explorer in visual studio and check binding redirects
> 
> binding redirects should use higher version

Thanks. Ok I will try that.

PR #302, opened 2024-10-16
fix: Avoid deprecation warnings in CMakeLists.txt by requiring minimum version 3.5

simonhyll:
### 🤔 What's changed?

CMake below 3.5 is going to be deprecated. Setting the minimum version to 3.5 suppresses that annoying warning. Also, the minimum version should be set before the project, otherwise you also get an annoying warning about that.

Warnings are annoying when they're this simple to fix.

### 📋 Checklist:

----

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -1,5 +1,5 @@
-PROJECT(gherkin C)
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.5)
+project(gherkin C)
 enable_testing()
 
 # Install library & headers in your directory


luke-hill:
I don't know enough about c to make an informed decision on this. But do you think this should be communicated in changelog @simonhyll 

Also is this a runtime change? If so should it go into a minor release? Or is this something that would have to be considered a breaking change.

@ursfassler are you still taking an interest in the c project?

ursfassler:
@luke-hill I am interested in having a working C/C++ cucumber implementation, but didn't knew/monitor this c code (I am on https://github.com/cucumber/cucumber-cpp/)
As I didn't had much time for it to, I am not up to date of the plans for the cucumber C/C++ implementation.

ursfassler:
CMake 3.5 seems already ancient.
The change should not have an impact on the code, but it could break a build (i.e. may need some manual changes to fix the build).

luke-hill:
I'll wait to merge until you have more understanding and agree this is the way forward. As the next release is just going to be 30.0.1

mpkorstanje:
So given that we currently don't have a dedicated maintainer for C and given that we do have a pull request that solve a problem for someone using Gherkin C, I think we can merge this.

PR #316, opened 2024-11-14
Fix missing gherkin-languages.json in Python wheel

musicinmybrain:
### 🤔 What's changed?

Ensures `gherkin/gherkin-languages.json` is included in Python wheels.

In a checkout, in `gherkin/python`

```
$ python3 -m build
$ python3 -m wheel unpack dist/gherkin_official-30.0.1-py3-none-any.whl
$ ls gherkin_official-30.0.1/gherkin/*.json
ls: cannot access 'gherkin_official-30.0.1/gherkin/*.json': No such file or directory
```

If we install the wheel in a virtualenv and try to `import gherkin.dialect`, the missing JSON file causes a `FileNotFoundError`.

I think this must be a regression since the [last successful PyPI release](https://pypi.org/project/gherkin-official/), currently lagging behind at 29.0.0, since 

### 📋 Checklist:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org).
 This document is formatted according to the principles of [Keep A CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Fixed
+- [Python] File gherkin-languages.json was not included as package data
 
 ## [30.0.1] - 2024-11-12
 ### Changed
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -37,6 +37,9 @@ Changelog = "https://github.com/cucumber/gherkin/releases"
 [tool.setuptools.packages.find]
 include = ["gherkin", "gherkin.*"]
 
+[tool.setuptools.package-data]
+gherkin = ["*.json"]
+
 [tool.mypy]
 disallow_untyped_defs = true
 packages = ["gherkin"]


musicinmybrain:
> I'll wait for the review from youtux, but please do add a changelog entry!

Ok, I will.

Now that 30.0.3 actually released on PyPI, I can demonstrate the problem there:

```
$ python3 -m venv _e
(_e) $ . _e/bin/activate
(_e) $ pip install gherkin-official
(_e) $ $ python -c 'import gherkin.dialect'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import gherkin.dialect
  File "/home/ben/src/forks/gherkin/_e/lib64/python3.13/site-packages/gherkin/dialect.py", line 31, in <module>
    with open(DIALECT_FILE_PATH, encoding="utf-8") as file:
         ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/ben/src/forks/gherkin/_e/lib64/python3.13/site-packages/gherkin/gherkin-languages.json'
```

youtux:
Oh that’s bad indeed. We should yank the release from PyPI @mpkorstanje .

We should also change our CI so that we test against the built package instead (as mentioned yesterday when fixing the release action)

mpkorstanje:
>  We should yank the release from PyPI

Done. New release made.

PR #318, opened 2024-11-14
python: Extract check build step

mpkorstanje:


.github/workflows/release-pypi.yaml
```diff
@@ -26,11 +26,12 @@ jobs:
       - name: Show Python version
         run: python --version
 
-      - name: Install Python package dependencies
+      - name: Check build
```
youtux, line 29: This is actually the build step, it should be named like “Build package”

PR #324, opened 2024-11-15
chore: update deps for Elixir, cucumber_messages 27.0.2

vitalis:


mpkorstanje:
Cheers. When making a pull request, it's useful to also update the `CHANGELOG`. It saves us from having to do it later. :laughing:

vitalis:
The CHANGELOG is not built automatically?

mpkorstanje:
Not yet. Would like to use conventional commits, but renovate can't detect and tag breaking changes correctly. So it's more reliable to determine the change type with the pull request.

All the guidelines were in the PR template.

vitalis:
NP, I'll update the PR

vitalis:
@mpkorstanje updated the PR

mpkorstanje:
Cheers. Apologies for the delay! It was rather busy around new years.

PR #328, opened 2024-11-23
[c, cpp]: support VERSION file for CMake ABI versioning

### LINKED ISSUE #320: Please version the ABI for C shared libraries
### 🤔 What's the problem you're trying to solve?

On Linux, system-wide shared libraries are expected to be versioned via `SONAME`.

### ✨ What's your proposed solution?

Set [`SOVERSION`](https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html) in `c/CMakeLists.txt`; this can be as simple as:

```cmake
set_target_properties(gherkin PROPERTIES SOVERSION 1)
```

Commit to incrementing the value of `SOVERSION` each time there is an ABI-incompatible change.

### ⛏ Have you considered any alternatives or workarounds?

[Downstream .so versioning](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_downstream_so_name_versioning) is an option.

### 📚 Any additional context?

It looks like the C++ shared library already does this correctly:

https://github.com/cucumber/gherkin/blob/f99769befdec0c3496027a2d09b716e6ab0c2f45/cpp/src/lib/gherkin/CMakeLists.txt#L34-L42

chybz:
### 🤔 What's changed?

- [c] slight update to existing CMakeFiles.txt to propagate VERSION.  Closes: #320 
- [cpp] add generic support for ABI versioning with VERSION

make sure ABI is versioned for shared libraries

### 📋 Checklist:

----


cpp/gherkin-cpp-rule-type.razor
```diff
@@ -22,7 +22,7 @@ public static string NameOf(Rule rule)
 #include <ostream>
 #include <string_view>
 
-namespace gherkin {
+namespace cucumber::gherkin {
```
mpkorstanje, line 25: Is this a breaking change?

chybz:
@mpkorstanje I'm going to merge. Can you please activate the `VERSION` automatic update for `gherkin/cpp` and `messages/cpp` ?

mpkorstanje:
It's already merged. https://github.com/cucumber/polyglot-release/pull/104

Question though, is the change in namespace that you included breaking in a Semver context? Doesn't matter for merging but does matter when it comes to deciding the next version number.

chybz:
TLDR: yes. Sincerely hope that's not overly problematic.

It is from a pure compilation standpoint, but function signatures are exactly the same as before, so this should be a minimal hassle for the (no more than 2 people I think) current cpp user-base.

mpkorstanje:
Okay cheers! I didn't see your edit until now.

mpkorstanje:
And no problem!

PR #346, opened 2025-01-03
[c]: install headers under cucumber/gherkin/ and .cmake files under gherkin/

musicinmybrain:
Rather than installing all the C headers directly under e.g. `/usr/include`, put them in a subdirectory like `/usr/include/cucumber/gherkin`.

This is *not* the same directory as the C++ headers, which have an extra `cucumber/` corresponding to the `cucumber::` C++ namespace, e.g. `/usr/include/cucumber/cucumber/gherkin/`. See https://github.com/cucumber/messages/issues/267#issuecomment-2478224301.

----

Rather than installing `.cmake` files directly under e.g. `/usr/lib/cmake`, put them in a subdirectory like `/usr/lib/cmake/gherkin`.

C++ CMake files are currently installed under e.g. `/usr/lib/cmake/cucumber_gherkin`; I think the leading `cucumber_` corresponds to the `cucumber::` C++ namespace, which is why I have chosen just `gherkin` for the C CMake files.

These paths will be influenced by https://github.com/cucumber/gherkin/pull/345.

### 🤔 What's changed?

Install headers under the subdirectory `cucumber/gherkin` within the target include directory rather than all at the top level. Install `.cmake` files under the `gherkin` subdirectory within the target `cmake` directory rather than all at the top level, which is more conventional.

The C API has a large number of header files with generic names like `event.h` and `item.h` that have a high risk of conflict in system-wide installations and therefore should not be installed directly in directories like `/usr/include`.

See https://github.com/cucumber/messages/issues/267#issuecomment-2478224301 for discussion of the path scheme.

A review of `-devel` packages installed on my Fedora system showed that installing `.cmake` files in a subdirectory of a `cmake` directory as in this PR is much more conventional than installing all the `.cmake` files at the top level.

Confirm that the adjusted installation paths are the desired ones.

### 📋 Checklist:

- [ ] My change requires a change to the documentation. **???**

----


mpkorstanje:
> Confirm that the adjusted installation paths are the desired ones.

I can't do this.  @chybz can I leave this PR to you?

luke-hill:
ping @chybz we're doing a big update to gherkin so feels good to get this in if we can. WDYT>?

musicinmybrain:
@chybz, any thoughts on this?

I’m preparing to submit a `gherkin` package for Fedora that will offer the Python implementation and (now that https://github.com/cucumber/gherkin/pull/376 has been merged and released) the C++ implementation and command-line tools. I’d like to enable the C implementation too, but I don’t want to do so until I’m sure I won’t have to move the files around later.

Thanks!

mpkorstanje:
@musicinmybrain can you rebase with main and see if the acceptance tests pass?

PR #347, opened 2025-01-04
Format generated Python parser module

kieran-ryan:
### 🤔 What's changed?

- Updated parser generation code with Python formatting as per style guidelines
- Invoked no-operation `Token.detach` method in generated parser code
- Migrated Python formatter from [`black`](https://black.readthedocs.io/en/stable/) to [`ruff`](https://docs.astral.sh/ruff/formatter/)
- Applied refactorings to generated parser code
  - Reorganised imports - [I001](https://docs.astral.sh/ruff/rules/unsorted-imports/) (`ruff check gherkin/parser.py --select=I001`)
  - Dropped `or False` logic from conditionals - [RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/) (`ruff check gherkin/parser.py --select=RET505`)
  - Simplified `ast_builder` instance assignment
    - `self.ast_builder = ast_builder if ast_builder is not None else AstBuilder()` -> `self.ast_builder = ast_builder or AstBuilder()`
  - Switched series of `or` statements to `any` call

- Ensures entire Python implementation is formatted consistently
  - Improves code review as pull requests less-susceptible to stylistic changes
  - Simplifies configuration - removes exclusion of incompliant modules
  - Improves readability
    - Bulk of change reduces mix of 4 and 8 space indentation within `match_token_at_'x'` methods to just 4 - which had been difficult to read and was most-likely a workaround in the generator code to satisfy both cases of running inside conditionals (with 4 spaces) and outside conditionals (8 spaces) without handling appropriately; a `SyntaxError` would be raised if standardised to 4 spaces without handling (as per this pull request)
- Migrating formatting to ruff has a number of advantages - including performance, and enabling standardising to a single toolset for linting and formatting
  - `ruff` is monorepo-friendly, with [hierarchical and cascading configuration](https://docs.astral.sh/ruff/configuration/#config-file-discovery) - which is ideal for this repository: allowing configuration to be stored within the `python` directory while allowing pre-commit, IDEs and commands run from the root of the repository
  - Aligns formatting development experience with pytest-bdd (pytest-dev/pytest-bdd#758)
- General refactorings to improve code quality and simplify code generation
  - `or False` is redundant
  - Easier to generate `any` calls against a comma-separate iterable compared to a series of `or` statements where it can't be applied on every line
  - _superflous_ `else` is redundant ([superflous-else-return - RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/))
  - Generated parser code currently contains a `useless-expression` ([B018](https://docs.astral.sh/ruff/rules/useless-expression/)) `token.detach` - updated code invokes appropriately though this doesn't actually alter behaviour (being a no-operation method) though conforms to intention and other parser implementation

- To evaluate changes are formatted correctly:

    Open a codespace (or locally alternatively):

    [Open in GitHub Codespaces](https://codespaces.new/cucumber/gherkin/tree/debt/format-py-parser)

    Regenerate all parser implementations:

    ```console
    docker build --tag berp-env .
    docker run --rm --interactive --tty --volume ".:/app" berp-env
    make clean-generate
    make generate
    ```

    From a separate terminal window, install the pre-commit hook and validate all Python code is formatted (including the generated parser code):

    ```console
    pip install pre-commit
    pre-commit install
    pre-commit run --all-files ruff-format
    ```

    Run Python unit and acceptance tests to validate no regression in the generated parser implementation:

    ```console
    pip install pytest
    pytest
    cd python/
    make acceptance
    ```

- Skipped formatting on `state_comment` string
  - Challenging to generate formatted Python multiline string from C# - consider exclusion pragmatic

- Intend to update GitHub Actions workflow in a follow up pull request to validate, or fully assess use of [pre-commit-ci](https://pre-commit.ci/) with this polyglot repo

### 📋 Checklist:


youtux:
I’m 👎 for this change:
> self.ast_builder = ast_builder if ast_builder is not None else AstBuilder() -> self.ast_builder = ast_builder or AstBuilder()

the reason is that you should be able to subclass AstParser and override `__bool__` to return False, and the above code should still use your object.

python/gherkin/parser.py
```diff
@@ -235,3090 +281,3753 @@ def match_token(self, state: int, token: Token, context: ParserContext) -> int:
...
-                return 0
-
-        state_comment = "State: 0 - Start"
-        token.detach
-        expected_tokens = ["#EOF", "#Language", "#TagLine", "#FeatureLine", "#Comment", "#Empty"]
-        error = UnexpectedEOFException(token, expected_tokens, state_comment) if token.eof() else UnexpectedTokenException(token, expected_tokens, state_comment)
+            self.build(context, token)
+            return 0
+
+        state_comment = "State: 0 - Start"  # fmt: skip
```
youtux, line 318: why is this `# fmt: skip` necessary?

kieran-ryan: > - Have skipped formatting on state_comment string
>   - Challenging to generate formatted Python multiline string from C# - consider exclusion pragmatic

Would be this comment; many of the generated `state_comment` variables exceed 88 characters (defaults for `ruff` and `black`), so in some cases would format as single-line strings and other cases multi-line. Unsure whether there would be a reliable, clean way of generating both from the `.razor` module that is less susceptible to upcoming changes (e.g. astral-sh/ruff#6936) - and observing the string formatting for both `black` and `ruff` has had issues or been difficult to ascertain direction. Approach I would probably take if implementing would be to evaluate comment length and use a multi-line string when appropriate accordingly.

Happy to go with your preference or guidance on this one.

<details>
<summary>`git diff` following `ruff format python/gherkin/parser.py` without `# fmt: skip`</summary>

```diff
@kieran-ryan ➜ /workspaces/gherkin (debt/format-py-parser) $ git diff
diff --git a/python/gherkin/parser.py b/python/gherkin/parser.py
index b74656b2..71fb03f0 100644
--- a/python/gherkin/parser.py
+++ b/python/gherkin/parser.py
@@ -351,7 +351,9 @@ class Parser:
             self.build(context, token)
             return 1
 
-        state_comment = "State: 1 - GherkinDocument:0>Feature:0>FeatureHeader:0>#Language:0"
+        state_comment = (
+            "State: 1 - GherkinDocument:0>Feature:0>FeatureHeader:0>#Language:0"
+        )
         token.detach()
         expected_tokens = [
             "#TagLine",
@@ -385,7 +387,9 @@ class Parser:
             self.build(context, token)
             return 2
 
-        state_comment = "State: 2 - GherkinDocument:0>Feature:0>FeatureHeader:1>Tags:0>#TagLine:0"
+        state_comment = (
+            "State: 2 - GherkinDocument:0>Feature:0>FeatureHeader:1>Tags:0>#TagLine:0"
+        )
         token.detach()
         expected_tokens = [
             "#TagLine",
@@ -452,7 +456,9 @@ class Parser:
             self.build(context, token)
             return 4
 
-        state_comment = "State: 3 - GherkinDocument:0>Feature:0>FeatureHeader:2>#FeatureLine:0"
+        state_comment = (
+            "State: 3 - GherkinDocument:0>Feature:0>FeatureHeader:2>#FeatureLine:0"
+        )
         token.detach()
         expected_tokens = [
             "#EOF",
@@ -661,7 +667,9 @@ class Parser:
             self.build(context, token)
             return 7
 
-        state_comment = "State: 6 - GherkinDocument:0>Feature:1>Background:0>#BackgroundLine:0"
+        state_comment = (
+            "State: 6 - GherkinDocument:0>Feature:1>Background:0>#BackgroundLine:0"
+        )
         token.detach()
         expected_tokens = [
             "#EOF",
@@ -878,7 +886,9 @@ class Parser:
             self.build(context, token)
             return 9
 
-        state_comment = "State: 9 - GherkinDocument:0>Feature:1>Background:2>Step:0>#StepLine:0"
+        state_comment = (
+            "State: 9 - GherkinDocument:0>Feature:1>Background:2>Step:0>#StepLine:0"
+        )
         token.detach()
         expected_tokens = [
             "#EOF",
@@ -2004,7 +2014,9 @@ class Parser:
             self.build(context, token)
             return 24
 
-        state_comment = "State: 23 - GherkinDocument:0>Feature:3>Rule:0>RuleHeader:1>#RuleLine:0"
+        state_comment = (
+            "State: 23 - GherkinDocument:0>Feature:3>Rule:0>RuleHeader:1>#RuleLine:0"
+        )
         token.detach()
         expected_tokens = [
             "#EOF",
```
</details>

youtux: Ah thank you for the explanation!

kieran-ryan:
> I’m 👎 for this change:
> > self.ast_builder = ast_builder if ast_builder is not None else AstBuilder() -> self.ast_builder = ast_builder or AstBuilder()
> 
> the reason is that you should be able to subclass AstParser and override `__bool__` to return False, and the above code should still use your object.

Makes sense - applied! Had been unsure on the same for a somewhat related reason - being too liberal by assessing "truthiness" rather than the explicit type declarations - which could become a consideration with changes down the line. Accustomed to using in a local context without this consideration. Interesting, thank you.

python/gherkin/parser.py
```diff
@@ -235,3090 +281,3753 @@ def match_token(self, state: int, token: Token, context: ParserContext) -> int:
...
+            UnexpectedEOFException(token, expected_tokens, state_comment)
+            if token.eof()
+            else UnexpectedTokenException(token, expected_tokens, state_comment)
+        )
         if self.stop_at_first_error:
             raise error
         self.add_error(context, error)
         return 48
+
     # GherkinD
```
youtux, line 4018: Previously we would short-circuit the evaluation at the first expression that was true, but now we always evaluate all the expressions.

is it possible this brings a significant performance decrease while parsing?

kieran-ryan: Will check this out - thanks for pointer

luke-hill:
Given this is a big refactor how far down the line is it. Is it close to being merge-worthy? @kieran-ryan as we're doing a lot of other big changes in other flavours so would be good to get this in if we can

python/gherkin-python.razor
```diff
@@ -157,39 +175,56 @@ class @(Model.ParserClassName):
...
     def lookahead_@(lookAheadHint.Id)(self, context: ParserContext, currentToken: Token) -> bool:
-        currentToken.detach
+        currentToken.detach()
         token = None
         queue = []
         match = False
         while True:
             token = self.read_token(context)
-            token.detach
+            token.detach()
```
youtux, line 207: interesting. we were not calling these methods, and still everything worked?

youtux: ah, I see now that the implemetation is a noop

PR #282, opened 2024-09-15
Drop python 2 compatibility

youtux:
### 🤔 What's changed?

Dropped python 2 compatibility.

### ⚡️ What's your motivation? 

I want to improve the python codebase, initially adding type annotations. But the python2 compatibility code makes it harder, so since python2 was EOL'd 4 years ago, I'd propose to remove supporting it.

Users needing python2 can still use older versions of the package.

### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)
- :boom: Breaking change (incompatible changes to the API)

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)
- [ ] I've changed the behaviour of the code
  - [ ] I have added/updated tests to cover my changes.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
- [x] Users should know about my change
  - [x] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

PR #352, opened 2025-01-08
Expose Python API as package imports

kieran-ryan:
### 🤔 What's changed?

- Exposed Python API as package imports

```diff
- from gherkin.parser import Parser
- from gherkin.pickles.compiler import Compiler
+ from gherkin import Compiler, Parser
```

- Dropped `count_symbols_test.py` unit test module

- Simplifies API and package usage - such as IDE completions for the public API
- Aligns with the [JavaScript API](https://github.com/cucumber/gherkin/blob/4412fb92486480bde8a31bc7d48ecd1e8fb164c6/javascript/src/index.ts)
- `count_symbols_test.py` module tested Python compatibility code; which has been removed (#282)

- Whether anything should additionally be exposed; or indeed reverted

### 📋 Checklist:


python/gherkin/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+from .ast_builder import AstBuilder
+from .dialect import DIALECTS, Dialect
+from .parser import Parser
+from .pickles.compiler import Compiler
+from .stream.source_events import source_event
+from .token_matcher import TokenMatcher
+from .token_matcher_markdown import GherkinInMarkdownTokenMatcher
+from .token_scanner import TokenScanner
+
```
jsa34, line 13: Maybe it's just me, but I usually order these alphabetically?

python/gherkin/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    "DIALECTS",
+    "AstBuilder",
+    "Compiler",
+    "Dialect",
+    "GherkinInMarkdownTokenMatcher",
+    "Parser",
+    "TokenMatcher",
+    "TokenScanner",
+    "source_event",
+]
```
youtux, line 24: I'd just expose `Compiler` and `Parser`, since are the only ones we "document".
If downstream needs the others, they can import their fully qualified name.

kieran-ryan: Sounds good yep - updated 👍

youtux (approved):

kieran-ryan:
Thank @youtux for review - and similarly @jsa34 (saw you left and dropped comment) - appreciate it both!

jsa34:
I dropped it because I thought I was being too picky! 🫣

kieran-ryan:
> I dropped it because I thought I was being too picky ( apologies! 🫣)

I actually would have made same comment @jsa34! 😄 Appreciate raising! Is interesting. The way I would have formatted `__all__` was as you described - alphabetically; though with ruff's automatic fix for sorting `__all__` - [RUF022 (unsorted-dunder-all)](https://docs.astral.sh/ruff/rules/unsorted-dunder-all/) - it performs a [natural sort](https://en.wikipedia.org/wiki/Natural_sort_order), respecting case first before alphabetic ordering. Something we can format and validate - which is cool; though perhaps not the 'go to' style for many.

```python
__all__ = [
    "a_variable",
    "CapWords",
    "UPPER_CASE",
]
```

```console
ruff check --select=RUF022 module.py --fix
```

```python
__all__ = [
    "UPPER_CASE",
    "CapWords",
    "a_variable",
]
```

jsa34:
Ah!  Never knew - every day's a school day!  Thanks for the detailed reply!

PR #344, opened 2025-01-03
[.NET] Avoid allocation and improve parsing time

obligaron:
### 🤔 What's changed?

This PR optimises allocations and speed by
- changing some types from class to struct
- removed/deferred some string trimming in `GherkinLine`
- change some yield return methods to explicit struct enumerators (no allocations & can be inlined)
- replaced arrays with `IEnumerable` so that `List` can be used, this avoids re-creating and coping with explicitly sized arrays

Results before:
| Method      | Runtime              | FeatureFile       | Mean      | Error     | StdDev    | Gen0     | Gen1     | Allocated  |
|------------ |--------------------- |------------------ |----------:|----------:|----------:|---------:|---------:|-----------:|
| **Parser**      | **.NET 8.0**             | **tags.feature**      |  **23.01 μs** |  **0.440 μs** |  **0.556 μs** |   **2.9297** |   **0.1221** |   **49.04 KB** |
| ParserReuse | .NET 8.0             | tags.feature      |  15.49 μs |  0.215 μs |  0.201 μs |   2.5024 |   0.0916 |   41.18 KB |
| Parser      | .NET Framework 4.8.1 | tags.feature      |  31.99 μs |  0.613 μs |  0.656 μs |   9.2773 |   0.4272 |   57.32 KB |
| ParserReuse | .NET Framework 4.8.1 | tags.feature      |  23.45 μs |  0.105 μs |  0.088 μs |   7.3853 |   0.2747 |   45.45 KB |
| **Parser**      | **.NET 8.0**             | **very_long.feature** | **565.50 μs** |  **5.492 μs** |  **5.138 μs** |  **95.7031** |  **54.6875** | **1568.74 KB** |
| ParserReuse | .NET 8.0             | very_long.feature | 546.43 μs | 10.785 μs | 10.088 μs |  94.7266 |  47.8516 | 1558.77 KB |
| Parser      | .NET Framework 4.8.1 | very_long.feature | 931.86 μs |  5.545 μs |  5.187 μs | 261.7188 |  96.6797 | 1610.54 KB |
| ParserReuse | .NET Framework 4.8.1 | very_long.feature | 896.85 μs |  2.761 μs |  2.447 μs | 259.7656 | 100.5859 | 1596.55 KB |

Results after:
| Method      | Runtime              | FeatureFile       | Mean       | Error     | StdDev    | Gen0     | Gen1    | Allocated |
|------------ |--------------------- |------------------ |-----------:|----------:|----------:|---------:|--------:|----------:|
[...]

PR #356, opened 2025-01-15
Additional error handling edge cases

obligaron:
### 🤔 What's changed?

- Added additional error handling edge cases as bad test data
- [c] Optimise error handling for empty datatable rows
- [Perl] Optimise error handling for unclosed DocStrings

These error handling edge cases were reported to the .NET implementation and fixed in an upstream project in 2019. Later, these fixes were also merged into the gherkin repository.
While working on some optimisations (see #344), the question arose whether the regression tests should be ported to the gherkin repository. This ensures that all implementations can handle these edge cases that have been reported by users.

The Perl implementation in general (first time writing Perl 😉), and also whether the required check for EOF should be added to other cases in the Perl implementation (but maybe this is something for another PR).

### 📋 Checklist:

----

### DIFF
--- a/perl/lib/Gherkin/TokenMatcher.pm
+++ b/perl/lib/Gherkin/TokenMatcher.pm
@@ -238,6 +238,9 @@ sub match_StepLine {
 
 sub match_DocStringSeparator {
     my ( $self, $token ) = @_;
+    if ($token->is_eof) {
+        return 0;
+    }
     if ( !$self->_active_doc_string_separator ) {
         return $self->_match_DocStringSeparator( $token, '"""', 1 )
           || $self->_match_DocStringSeparator( $token, '```', 1 );
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 
 ### Fixed
 - [.NET] Fix NuGet package generation
+- [c] Optimise error handling for empty datatable rows
 
 ## [31.0.0] - 2025-01-29
 ### Added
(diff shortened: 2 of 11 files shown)


perl/lib/Gherkin/TokenMatcher.pm
```diff
@@ -238,6 +238,9 @@ sub match_StepLine {
 
 sub match_DocStringSeparator {
     my ( $self, $token ) = @_;
+    if ($token->is_eof) {
+        return;
```
mpkorstanje, line 242: Other implementations return `true`/`false` in `_match_DocStringSeparator`, which I think should cover this case.

To make it easier for everyone to compare across languages, it would be good to be consistent.

mpkorstanje (changes requested):
Looks good! 

Small request for Perl. Do you think you can do that? If not, I can take a stab at it.

perl/lib/Gherkin/TokenMatcher.pm
obligaron: I added an explicit false/0

testdata/bad/backslash_at_end_of_line_in_DataTable.feature
```diff
@@ -0,0 +1,5 @@
+Feature: Addition
+Scenario: Add two numbers
+	When I press
+		| foo |
+		| bar \
```
luke-hill, line 5: Think this file should be snake cased and there is a newline missing here

obligaron: I have changed all the testdata file names.

Regarding the missing newline: I think this is part of the tests because they should resemble someone editing the feature files in an incomplete state. The gherkin parser should parse them. So I think this is OK in this case?

I copied the tests from the [reqnroll ide plugin](https://github.com/reqnroll/Reqnroll.VisualStudio/blob/main/Tests/Reqnroll.VisualStudio.Tests/Editor/Services/DeveroomGherkinParserTests.cs).

PR #372, opened 2025-02-11
java: Optimize GherkinLine performance

### LINKED ISSUE #361: Performance issue in StringUtils
### 👓 What did you see?

In the `Parser.parse()` method, 30% to 50% of the elapse time is passed doing `String` trimming in the `StringUtils` class, see https://github.com/cucumber/gherkin/issues/351#issuecomment-2627758354

### ✅ What did you expect to see?

No doubt `String` trimming is important, but it should not last that long.

### 📦 Which tool/library version are you using?

OpenJDK21
gherkin 31.0.0 (the main branch)

### 🔬 How could we reproduce it?

See JMH benchmark code in https://github.com/cucumber/gherkin/issues/351#issuecomment-2627758354

### 📚 Any additional context?

Identified for #351

jkronegg:
### 🤔 What's changed?

It started as a temptative to improve `StringUtils` (#361), but ended up with many other small improvements, all based on JMH micro-benchmark and IntelliJ profiler.

The following improvements have been done (no public API modified):
- `GherkinLine`: 
    - constructor/`getTableCells()`: rewrote trim+symbolCount logic to an integrated operation which trim and intent at the same time
    - `getTags()`: replaced split by String traversal and using compiled Regexp
- `GherkinDocumentBuilder`: compiled regexps and simplified mapping between `TokenType` and `RuleType`
- `GherkinDialect`: precomputing list size and removed duplicates
- `EncodingParser`: avoid split on the while file to split only the first lines

The parser is now 1.7x faster (=40%) on the `very_long.feature` test file, as reflected by JMH micro-benchmark:

     MyClassBenchmark.original  avgt   25  5514.929 ± 943.017  us/op
     MyClassBenchmark.modified  avgt   25  3265.629 ± 287.454  us/op

There is the JMH code to reproduce the results:
```
package io.cucumber.gherkin;

import io.cucumber.messages.types.Envelope;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;

import java.io.IOException;
import java.nio.file.Paths;
import java.util.concurrent.TimeUnit;

public class MyClassBenchmark {
    @Benchmark
    @BenchmarkMode(Mode.AverageTime)
    @OutputTimeUnit(TimeUnit.MICROSECONDS)
    public Stream<Envelope> original() throws IOException {
        return GherkinParser.builder().build().parse(Paths.get("../testdata/good/very_long.feature"));
    }
}
```

On a real project with 1000 scenarios, 50 parameterTypes and 250 step definitions, the IntelliJ profiler gives for `GherkinMessagesFeatureParser.parse`:
- original version (gherkin 31.0.0): 434 ms
- modified version (this PR): 209 ms

That's 2.1x faster... 😁

Fixes cucumber/gherkin#361 

On this PR, we can run the following test:

    @Test
    void test_for_profiler_parser() throws IOException {
        for (int i=0; i<1000; i++) GherkinParser.builder().build().parse( Paths.get("../testdata/good/very_long.feature"));
    }

Below is the Intellij profile flame graph for this test:
image

The is still some little room for improvement in:
- `getLocation` (8%): avoid using `Long` values in `cucumber-messages` when primitive types can be used, see https://github.com/cucumber/messages/issues/295
- `cucumber-messages` (1%): use Java 10+ to avoid recreating immutable lists, see https://github.com/cucumber/messages/issues/282

I'm not counting the `UUID.randomUUID()` because it can be easily solved by selecting a faster UUID generator (e.g. `IncrementingUuidGenerator`) by configuring Cucumber properly.

### 📋 Checklist:


luke-hill:
Given this is close to completion @jkronegg i'll wait til you fix up the last bits of the codegeneration task.

Quick question, is this really a bug fix or more of a generic change? Just noticed you popped a changelog in fixed?

jkronegg:
From my point of view, performance issues are bugs (given the non-functional requirement "the application must go as fast as possible"😁). But someone else could consider this PR as an improvement given no one asked for that NFR😅.
@luke-hill as you prefer.

luke-hill:
As/when it's reviewed by Rien he can best advise as this is a big Java improvement. I'm in no way able to advise on Java stuff

jkronegg:
I made some additional performance improvements, with the following JMH microbenchmark results: 

```
MyClassBenchmark.original  avgt   25  3430.073 ± 217.549  us/op
MyClassBenchmark.modified  avgt   25  1716.679 ± 15.788  us/op
```

So that's now 2 x faster. There is IntelliJ profiler flame graph:
image

Creating the `cucumber-messages` objects now take about 10.5% of the total time (see https://github.com/cucumber/messages/issues/283 and https://github.com/cucumber/messages/issues/282 ). Thus, there is still a little bit to improve on that point. But otherwise I don't think I can do really better...

mpkorstanje (changes requested):
Cheers. I'm about half way through this PR.

java/src/main/java/io/cucumber/gherkin/EncodingParser.java
```diff
@@ -41,15 +42,18 @@ private static byte[] removeByteOrderMarker(byte[] source) {
...
-        for (String line : source.split("[\\n\\r]")) {
-            if (!COMMENT_OR_EMPTY_LINE_PATTERN.matcher(line).find()) {
-                return Optional.empty();
-            }
-            Matcher matcher = ENCODING_PATTERN.matcher(line);
-            if (matcher.find()) {
-                String charSetName = matcher.group(1).toUpperCase(ROOT);
-                Charset charset = Charset.forName(charSetName);
-                return Optional.of(charset);
+        try (Scanner scanner = new Scanner(source)) {
```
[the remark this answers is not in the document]
jkronegg: I replaced the `Scanner` by a `Matcher.find()` which consume the feature file line by line instead of doing a `String.split()` on the full feature file.

mpkorstanje:
And reviewing this the second time, I've completely forgotten what I was looking at. In other words, this PR is too big. I'll be splitting into smaller parts. If you have time to do the same, that would be appreciated!

jkronegg:
> And reviewing this the second time, I've completely forgotten what I was looking at. In other words, this PR is too big. I'll be splitting into smaller parts. If you have time to do the same, that would be appreciated!

Sure, I'll do it.

java/src/main/java/io/cucumber/gherkin/GherkinDocumentBuilder.java
```diff
@@ -243,6 +235,19 @@ private Object getTransformedNode(AstNode node) {
         return node;
     }
 
+    private static String concatMatchedText(List<Token> lineTokens) {
+        // we guess that the average line is about 50 characters
+        StringBuilder content = new StringBuilder(50 * lineTokens.size());
```
mpkorstanje, line 240: In `InputStreams` the assumption was 100 characters per line.

java/src/main/java/io/cucumber/gherkin/GherkinDocumentBuilder.java
```diff
@@ -163,14 +158,11 @@ private Object getTransformedNode(AstNode node) {
                 List<Token> lineTokens = node.getTokens(TokenType.Other);
                 // Trim trailing empty lines
                 int end = lineTokens.size();
-                while (end > 0 && lineTokens.get(end - 1).matchedText.matches("\\s*")) {
+                while (end > 0 && PATTERN_SPACES.matcher(lineTokens.get(end - 1).matchedText).matches()) {
```
mpkorstanje, line 161:
```suggestion
                while (end > 0 && lineTokens.get(end - 1).line.isEmpty()) {
```

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -65,20 +63,26 @@ public String getRestTrimmed(int length) {
 
     @Override
     public List<GherkinLineSpan> getTags() {
-
-        String uncommentedLine = trimmedLineText.split("\\s" + COMMENT_PREFIX, 2)[0];
-        List<GherkinLineSpan> tags = new ArrayList<>();
+        // in most cases, the line contains no tag, so the code is optimized for this situation
+        if (emptyTrimmedLineText) {
```
mpkorstanje, line 67: Can also be applied to `getTableCells`.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -124,10 +128,9 @@ public List<GherkinLineSpan> getTableCells() {
                         // Skip the first empty span
                         beforeFirst = false;
                     } else {
-                        String cell = cellBuilder.toString();
-                        String leftTrimmedCell = ltrimKeepNewLines(cell);
-                        int cellIndent = symbolCount(cell) - symbolCount(leftTrimmedCell);
-                        lineSpans.add(new GherkinLineSpan(cellStart + cellIndent + OFFSET, rtrimKeepNewLines(leftTrimmedCell)));
+                        GherkinLineSpan gherkinLineSpan = new GherkinLineSpan(cellStart + OFFSET);
+                        StringUtils.trimAndIndentKeepNewLines(cellBuilder.toString(), gherkinLineSpan);
```
mpkorstanje, line 132: This is confusing, through `Indentable` the `GherkinLineSpan` has its text set.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -148,4 +151,10 @@ public boolean startsWithTitleKeyword(String text) {
                 trimmedLineText.startsWith(GherkinLanguageConstants.TITLE_KEYWORD_SEPARATOR, textLength);
     }
 
+    @Override
+    public void indent(int indent, String trimmedText) {
+        this.indent = indent;
```
mpkorstanje, line 156: This uses `+=` in `GherkinLineSpan`.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
mpkorstanje: Something like 
```
gherkinLineSpan =  trimAndIndentKeepNewLines(content, (trimmedContent, trimmedColumns) -> new GherkinLineSpan(trimmedContent, cellStart + trimmedColumns + OFFSET)
```

java/src/main/java/io/cucumber/gherkin/GherkinDocumentBuilder.java
```diff
@@ -22,13 +22,26 @@
...
 
 class GherkinDocumentBuilder implements Builder<GherkinDocument> {
+    public static final Pattern PATTERN_SPACES = Pattern.compile("\\s*");
+    /**
+     * Cache of Long objects for the range 0-3999. This is used
+     * to avoid creating a huge amount of Long objects in getLocation().
+     * We can't use Long.valueOf() because it caches only the first 128
+     * values, and typical feature files have much more lines.
+     */
+    private final static Long[] longs = new Long[4000];
```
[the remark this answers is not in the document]
jkronegg: see #387

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
jkronegg: No, because `getTableCells` is called by `TokenMatcher.match_TableRow()` and only if there is a table separator. Thus the line is never empty.

mpkorstanje:
Merged #387. Cheers! 

The only thing I don't understand is why the cache helps. 

The `TokenScanner` creates a `Location` for each line. And the `Long` used for that line number get reused. So we never create more longs than there are lines in a feature file. I also would not expect a feature file to be wider than 128 characters so `Locations.atColumn` shouldn't be slow either.

java/src/main/java/io/cucumber/gherkin/TokenMatcher.java
```diff
@@ -202,13 +208,24 @@ private StepKeywordType getKeywordType(String stepKeyword) {
...
         return false;
     }
 
+    private String removeDocStringIndent(Token token) {
+        if (activeDocStringSeparator == null) {
+            return token.line.getRawText();
+        }
+        if (indentToRemove > token.line.getIndent()) {
+            // BUG: Removes trailing whitespace!
+            return token.line.getText();
```
mpkorstanje, line 224: This bug can be sorted out later.

mpkorstanje (commented):
Cheers!  With this pull request and the spin-offs (#388, #385, #383, #382 #380), running the `test_for_profiler_parser` shows some serious improvement!

Before: 2.879s
After: 1.466

So that is almost a 1,96x speed up. Not as good as the 2.1x speed up you mentioned but probably within the margins of error.

jkronegg:
> The only thing I don't understand is why the cache helps.

The cache helps when you have more than one feature file. Without cache, as feature files have typically more than 128 lines, the `Long.valueOf` will create new Long instances for each feature file.

jkronegg:
> So that is almost a 1,96x speed up. Not as good as the 2.1x speed up you mentioned but probably within the margins of error given the testing method. If you see more areas for improvement I'd be happy to consider those!

The proposed `Indentable` interface was used to let the `StringUtils.trimAndIndent` method to set two values at the same time. Returning a pair of objects requires a new object instantiation, which requires more memory and cpu than setting two field values. I will do some comparisons, but my guess is that the performance difference comes from this modification.

mpkorstanje:
Ah. I see, I couldn't get my lambda based proposal to work easily.

Also I missed an unboxing to `Character`.

https://github.com/cucumber/gherkin/commit/0003000fe0f22407e0f85646f0df0c344bcbcd15

mpkorstanje:
I also think `TokenScanner` would be a good target for optimization. The `BufferedReader` seems unnecessary as the  `GherkinParser` aways provides a String as input

mpkorstanje:
Pushed e8513d450a7baefbaa2655df4d192d6ac438903b to make it easier to optimize the `TokenScanner`. I'm guessing a custom iterator that splits on `\r\n` would be quite efficient.

jkronegg:
> Ah. I see, I couldn't get my lambda based proposal to work easily.

I made a lambda based version, but performance is about 20% lower than the `Indentable` version. There's still some work on that topic...

mpkorstanje:
Oef. Might be worth looking at other parts (of the stack) first then.

Issue #373, opened 2025-02-16
[Python] Regression: Descriptions consisting of just a comment in a Scenario or Scenario Outline block throws error


jsa34:
### 👓 What did you see?

This commit: https://github.com/cucumber/gherkin/pull/334
Appears to have broken a number of unit tests for `pytest-bdd` due to the gherkin parser throwing errors on parsing feature files.
The cause is that the parser throws an error if the description if just a comment inside a Scenario or Scenario Outline.  It appears this was a missed test case scenario in the PR.

See CI job https://github.com/pytest-dev/pytest-bdd/actions/runs/13252365760/job/36993581825?pr=765 for errors.

Backtrace example:

```
test_scenario_comments.py:5: in <module>
    @scenario("comments.feature", "Comments")
/pytest-bdd/src/pytest_bdd/parser.py:480: in parse
    gherkin_doc: GherkinDocument = self._parse_feature_file()
/pytest-bdd/src/pytest_bdd/parser.py:476: in _parse_feature_file
    return get_gherkin_document(self.abs_filename, self.encoding)
/pytest-bdd/src/pytest_bdd/gherkin_parser.py:316: in get_gherkin_document
    gherkin_data = Parser().parse(feature_file_text)
/.venv/lib/python3.10/site-packages/gherkin/parser.py:96: in parse
    state = self.match_token(state, token, context)
/.venv/lib/python3.10/site-packages/gherkin/parser.py:231: in match_token
    return state_map[state](token, context)
/.venv/lib/python3.10/site-packages/gherkin/parser.py:795: in match_token_at_11
    self.end_rule(context, 'Description')
/.venv/lib/python3.10/site-packages/gherkin/parser.py:120: in end_rule
    self.handle_ast_error(context, rule_type, self.ast_builder.end_rule)
/.venv/lib/python3.10/site-packages/gherkin/parser.py:2800: in handle_ast_error
    self.handle_external_error(context, True, argument, action)
/.venv/lib/python3.10/site-packages/gherkin/parser.py:2813: in handle_external_error
    return action(argument)
/.venv/lib/python3.10/site-packages/gherkin/ast_builder.py:54: in end_rule
    self.current_node.add(node.rule_type, self.transform_node(node))
/.venv/lib/python3.10/site-packages/gherkin/ast_builder.py:272: in transform_node
    last_non_empty = next(
E   StopIteration
```

Feature file examples:
```
Feature: Comments
  Scenario: Comments
    # Comment
[...]

PR #334, opened 2024-12-09
 Allow comment inside descriptions

jacob-wieland-gematik:
### ⚡️ What's your motivation? 

Fixes: #333

### 🏷️ What kind of change is this?

- :zap: New feature (non-breaking change which adds new behaviour)

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)
- [x] I've changed the behaviour of the code
  - [x] I have added/updated tests to cover my changes.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

PR #378, opened 2025-03-07
Add test cases when there is only a comment between the scenario line and the first step

gasparnagy:
### 🤔 What's changed?

Added two more test cases according to #373 

Better understand #373 and see impact of #334

### 📋 Checklist:

----


mpkorstanje:
Looks like the AST builder is making assumptions it shouldn't:

```
self.current_node.add(node.rule_type, self.transform_node(node))
  File "/home/runner/work/gherkin/gherkin/python/gherkin/ast_builder.py", line 272, in transform_node
    last_non_empty = next(
StopIteration
```


@jsa34 @youtux would you be able to dig into this.

(Or anyone else with some Python knowledge)

jsa34:
I'll check out the code and debug :)

jsa34:
So, it looks like for v30.0.4 Gherkin, the AST builder didn't consider that there was a "Description" node when there is just a comment - it looks like it was discarded and so the "Description" was non-existent in this circumstance.
Since v31.0.0, the comment makes it think there is a Description AST node (wrongly or rightly - I'm not sure what is intended behaviour in the test scenarios above (i.e. description only with comment(s)).

My naive assessment just looking at the Python logic would be:
1. If we expect that the AST builder can have a Description node with an empty list of `line_token`s and this is valid, that we need to change the code to handle this - something along the lines of:
```
    line_tokens = node.get_tokens("Other")

    if not line_tokens:  # Handle the case where there are no tokens
        return ""

    # Trim trailing empty lines
    last_non_empty = next(
        (i for i, j in reversed(list(enumerate(line_tokens))) if j.matched_text),
        -1,  # Default value if no non-empty tokens are found
    )

    description = "\n".join(
        [token.matched_text for token in line_tokens[: last_non_empty + 1]]
    )

    return description
```
2. If we should only have a Description node if there are actually `line_token`s associated with it, this looks like a regression, as that was how the old logic seemed to work, and I'm not sure how to best resolve this.

I hope this is helpful!

mpkorstanje:
It would be advisable to use the other implementations as a reference. 

https://github.com/cucumber/gherkin/blob/1371351a7cf3c32d74c647ff02f75ae42fd197ae/python/gherkin/ast_builder.py#L269-L279

https://github.com/cucumber/gherkin/blob/1371351a7cf3c32d74c647ff02f75ae42fd197ae/javascript/src/AstBuilder.ts#L217-L227

https://github.com/cucumber/gherkin/blob/1371351a7cf3c32d74c647ff02f75ae42fd197ae/java/src/main/java/io/cucumber/gherkin/GherkinDocumentBuilder.java#L162-L174

Replacing the list comprehension with a simple while loop should solve the problems and keep the implementations aligned.

mpkorstanje:
So Gaspar made a branch directly in this repository. As a member of the Cucumber org you can push directly to that branch.

jsa34:
I updated the PR title to reflect that this now has the fix for the issue these tests exposed, too.
Thanks @gasparnagy for your help and thanks @mpkorstanje for your patience and support!

jsa34:
@mpkorstanje @gasparnagy are we happy with the changes?  It would be good to release a new version of gherkin when possible to fix this.

blaisep:
> Sorry, short on time again this week. At a glance it looks okay.
> 
> You should have the rights to merge this PR. And remember that neither me nor Gaspar are python experts so this is your call. If you do want feedback from other Python contributors you could ask them.
> 
Fwiw, @jsa34 's Python is generally exemplary, I'm unlikely to find room for improvement, and I am happy provide a review for the record.

PR #395, opened 2025-05-15
add a target `mostlyclean` in Makefiles

fperrad:
### 🤔 What's changed?

Add a new target `mostlyclean` which differs from `clean` by not removing the external dependencies.

`mostlyclean` is a well known name for a target, see <https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html>.

Speed up the development process with the mantra `make mostlyclean acceptance`
(external dependencies are seen as stable)

### 📋 Checklist:

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,7 +131,7 @@ good), but writing a few during development might help you progress.
 You'll spend quite a bit of time fiddling with the `.razor` template to make it
 generate code that is syntactically correct.
 
-When you get to that stage, `make clean acceptance` will run the acceptance tests, which iterate
+When you get to that stage, `make mostlyclean acceptance` will run the acceptance tests, which iterate
 over all the `.feature` files under `../testdata`, passes them through your
 `bin/gherkin-generate-tokens` and `bin/gherkin-generate-ast` command-line programs,
 and compares the output using `diff`.


cpp/Makefile
```diff
@@ -53,10 +53,13 @@ clean-gherkin-languages: ## Remove gherkin-languages.json and any derived files
 	echo "Nothing to do"
 .PHONY: clean-gherkin-languages
 
-clean: clean-deps ## Remove all build artifacts and files generated by the acceptance tests
+mostlyclean:
```
mpkorstanje, line 56: Does this need a documentation string for `make help`?

mpkorstanje:
Cheers! LGTM, small remark above

c/Makefile
```diff
@@ -56,9 +56,12 @@ clean-gherkin-languages: ## Remove gherkin-languages.json and any derived files
 	$(MAKE) --directory=src
 	touch $@
 
-clean:
+mostlyclean:
 	rm -rf .compared .built .run acceptance
-	$(MAKE) --directory=src $@
+	$(MAKE) --directory=src clean
```
mpkorstanje, line 61: Not every makefile further down has a `mostlyclean`, fixed it like this, but not sure if that is what you intended.

mpkorstanje:
Cheers!

PR #377, opened 2025-03-05
Markdown support for java

madhoks:
### 🤔 What's changed?

Adding Markdown support for Java. This will enable to define BDDs in Markdown which just simply render better.

### ⚡️ What's your motivation? 

Address #100 .
A new feature.
I love BDD, (well love may be a strong word, but ... you get the gist), and I love seeing BDDs in Markdown.
This was already supported in javascript so seemed simple enough to replicate to Java.

### 🏷️ What kind of change is this?

- :zap: New feature (non-breaking change which adds new behaviour)

### ♻️ Anything particular you want feedback on?

General code feedback.

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)
- [X] I've changed the behaviour of the code
  - [X] I have added/updated tests to cover my changes.
- [X] My change requires a change to the documentation.
  - [X] I have updated the documentation accordingly.
- [ ] Users should know about my change
  - [ X] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

PR #399, opened 2025-05-21
Markdown support for Perl

fperrad:
### 🤔 What's changed?

Adding **Markdown with Gherkin** (MDG) support for Perl.

Javascript is the reference implementation for MDG, but it lacks the `gherkin-generate-tokens` (and the corresponding tests).

So, I am not confident with the test suite.

Some points need clarification:

 - ~~with MDG, the messages `Comment` are skipped when building the AST : bug or feature ?~~ (feature)
 * [X] I try to reuse `testdata/good/*.feature.md.tokens` from https://github.com/cucumber/gherkin/pull/377, but there need some changes. So, these files need a review / validation.

### 📋 Checklist:

----


perl/lib/Gherkin/AstBuilder.pm
```diff
@@ -60,7 +60,7 @@ sub build {
             Cucumber::Messages::Comment->new(
                 location => $self->get_location($token),
                 text     => $token->matched_text
-            );
+            ) unless $self->{uri} =~ m/\.md$/;
```
mpkorstanje, line 63: This looks weird. Using Javascript as the reference implementation, I can't find this functionality.

mpkorstanje:
> with MDG, the messages Comment are skipped when building the AST : bug or feature ?

I'm not sure if I understand this. I don't think Markdown supports comments. What would a comment in Markdown look like?

> Javascript is the reference implementation for MDG, but it lacks the gherkin-generate-tokens (and the corresponding tests).

Would you be able to add the functionality to generate tokens to the JS implementation?

mpkorstanje (commented):
I've given this a once over. Other than my remark on the AST builder I don't see anything too odd. But this PR also exceeds my current capacity for in-depth review.

@ehuelsmann would you be so kind?

perl/lib/Gherkin/AstBuilder.pm
mpkorstanje: > I don't fully understand the Javascript implementation.

Well that makes two of us. :sweat: 

It looks intentional that table separators are matched as comments.

https://github.com/cucumber/gherkin/blob/214337104f9b110e5460c61528b760dfe93c5fbb/javascript/src/GherkinInMarkdownTokenMatcher.ts#L129C1-L136C4

But I can't work out how they're kept out of the AST.  And Unfortunately I don't have the time available for a real deep dive into this.  I hope Erik can provide some comments

perl/lib/Gherkin/MarkdownTokenMatcher.pm
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+sub match_Comment {
+    my ( $self, $token ) = @_;
+    if (   $token->line->startswith('|')
+        && $self->_is_gfm_table_separator( $token->line->table_cells ) )
+    {
+        $self->_set_token_matched( $token,
+            Comment => { text => $token->line->line_text, indent => 0 } );
```
ehuelsmann, line 217: @fperrad @mpkorstanje From what I understand from the JavaScript implementation, this is where the difference is: the JavaScript implementation sets a `token.matchedType` on all `match_XXX` functions *except* for on `match_Comment`. However, the Perl implementation *does* set it in this line by passing the "Comment =>" above. Try setting the `token.matchedType` to `TokenType.Comment` in the JS implementation and see if they now both fail the tests.

I *think* failing to set the `token.matchedType` simply causes comments to be ignored, which would be roughly the same as just not calling `_set_token_matched()` abov.

ehuelsmann (changes requested):
Yes. This is now the equivalent of the JS code. Very good.

While you were heads down implementing this MD support, the project received *and* fixed #400. Based on my review of the JS code and this PR both are lacking the required fix. Please add the fix and I'll PR for JS.

perl/lib/Gherkin/AstBuilder.pm
ehuelsmann: The naming of the function is misleading: Instead of matching it as a comment, it's actually matching it as "nothing at all", which is also why it's not ending up in the AST.

perl/lib/Gherkin/MarkdownTokenMatcher.pm
```diff
@@ -0,0 +1,345 @@
...
+sub new {
+    my ( $class, $options ) = @_;
+    $options->{'dialect'} ||= Gherkin::Dialect->new( { dialect => 'en' } );
+    my $self = bless $options, $class;
+    $self->_default_dialect_name( $self->dialect_name );
+    my @non_star_step_keywords = sort { length($b) <=> length($a) } map {
+        grep { $_ ne '* ' }
+          @{ $self->dialect->$_ }
+    } qw/Given When Then And But/;
+    $self->_non_star_step_keywords( \@non_star_step_keywords );
```
ehuelsmann, line 27: The values of the sorted keywords are dependent on the selected language, which means that it is better to move this bit (appending and sorting) to `change_dialect()`. Sorry for not noticing before.

BTW, I don't think that changing the language using the language tag is supported by GherkinInMarkdown, but at least putting it in `change_dialect()` will make the standard and markdown gherkin tokenizers behave the same way. (good for consistency; predictable behavior)

ehuelsmann: I hope you don't mind me addressing my own comment, but I've fixed this comment: I didn't want you to keep having to go back and forth over minor issues.

mpkorstanje:
I would say that more than a few days have passed. :wink:

PR #403, opened 2025-05-23
[PHP] Prefer the longest step keyword

stof:

Step keywords may be prefixes of each other. For example in French several variations of given start with `Etant donné`. When this is the case Gherkin should prefer the longest possible keyword.

This is currently only implicitly enforced by manually ordering the keywords.

However, for Creole language the prefixes are spread between then and when so manually sorting them is no longer possible.

This is the PHP implementation of the fix for #400.

### 📋 Checklist:

  - [x] I have added/updated tests to cover my changes. See #402

### DIFF
--- a/php/.gitignore
+++ b/php/.gitignore
@@ -8,3 +8,4 @@ composer.lock
 .deps
 .compared
 .tested
+.built
--- a/php/src/GherkinDialect.php
+++ b/php/src/GherkinDialect.php
@@ -21,6 +21,9 @@ namespace Cucumber\Gherkin;
  */
 final class GherkinDialect
 {
+    /** @var non-empty-list<non-empty-string> */
+    private readonly array $stepKeywords;
+
     /**
      * @param Dialect $dialect
      */
@@ -28,6 +31,7 @@ final class GherkinDialect
         private readonly string $language,
         private readonly array $dialect,
     ) {
+        $this->stepKeywords = $this->buildStepKeywords();
     }
 
     public function getLanguage(): string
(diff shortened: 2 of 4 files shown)


php/.gitignore
```diff
@@ -8,3 +8,4 @@ composer.lock
 .deps
 .compared
 .tested
+.built
```
stof, line 11: this `.built` file is an empty file used by Makefile as a marker. This avoids committing it by mistake when doing the commit without running `make mostlyclean` before (which would delete that file)

php/src/GherkinDialect.php
```diff
@@ -21,13 +21,17 @@
  */
 final class GherkinDialect
 {
+    /** @var non-empty-list<non-empty-string> */
+    private readonly array $stepKeywords;
```
stof, line 25: Previously, the merged list was built on each access (which means for each line of the file being parsed when attempting to match step lines). I'm now computing the merged list once and storing it, which is consistent with the way the Java implementation works.

mpkorstanje:
I've added you to the committers group. That should allow you to merge and run test pipelines on your own.  I trust you'll use this responsibly.

php/src/GherkinDialect.php
```diff
@@ -98,18 +102,29 @@ public function getButKeywords(): array
...
             ...$this->getButKeywords(),
         ];
-    }
 
-    /** @return non-empty-list<non-empty-string> */
-    public function getExamplesKeywords(): array
-    {
-        return $this->dialect['examples'];
+        // Sort from longest to shortest
+        usort($keywords, fn ($a, $b) => \strlen($b) - \strlen($a));
```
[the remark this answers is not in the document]
stof: I copied my code used to fix the in `behat/gherkin`, where this was implemented (by me as well) by thinking about the length of the keyword.
But it is true that we only care about the length when they have a common prefix, so `rsort` will do it.

PR #404, opened 2025-05-23
Prefer the longest keywords

tigh-latte:
Some langauges have two different step keywords starting with the same characters. In cases like these, we need to ensure that if a user types out the longer of these keywords, that is what gets matched.

This is done by sorting the keywords for each dialect.

As well as this, I've added handling to make sure that the sorting is only done once per dialect to make it inline with other fixes. This required a couple of fields to be added to the `Dialect` struct, so I've updated the `builtin_dialects.go.jq` file to built the `Dialect` struct in a way that allows zero base instantiation.

Golang implementation for #400 

I'd like an opinion on my approach to making sure the keyword list isn't sorted for every step.

I know I could have done a nil check, but that isn't thread safe and being honest I'm not sure how this interacts with running the godog suite concurrently, say, so I've took the threadsafe approach out of caution.

### 📋 Checklist:

----

### DIFF
--- a/go/dialect.go
+++ b/go/dialect.go
@@ -1,6 +1,11 @@
 package gherkin
 
-import messages "github.com/cucumber/messages/go/v24"
+import (
+	"sort"
+	"sync"
+
+	messages "github.com/cucumber/messages/go/v24"
+)
 
 type Dialect struct {
 	Language     string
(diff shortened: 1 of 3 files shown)


tigh-latte:
@mpkorstanje didn't even have to do it on the jq template, can just run the sort func in `init`.

mpkorstanje:
Cheers!

PR #408, opened 2025-05-25
[Perl] Prefer the longest step keyword

ehuelsmann:
### 🤔 What's changed?

Address the issue reported in #400 (Gherkin should prefer the longest prefix match on step keywords).

Advance all Gherkin implementations at the same time, including Perl.

N/A

### 📋 Checklist:

----


perl/lib/Gherkin/TokenMatcher.pm
```diff
@@ -213,7 +213,11 @@ sub _unescaped_docstring {
 
 sub match_StepLine {
     my ( $self, $token ) = @_;
-    my @keywords = map { @{ $self->dialect->$_ } } qw/Given When Then And But/;
+    my @keywords = sort {
```
mpkorstanje, line 216: Performance wise this is probably not a good idea.

mpkorstanje (commented):
There are more commits than I would expect in this pull request.

ehuelsmann:
> There are more commits than I would expect in this pull request.

I started out with `fix/prefer-longest-step-keyword`; originally there was only 1 commit (starting from "main"), but when I rebased, the extra commits came in. I'll be happy to put it back on "main" if that's the better option. I'll also address the performance review remark.

mpkorstanje:
> I started out with fix/prefer-longest-step-keyword; originally there was only 1 commit (starting from "main"), but when I rebased, the extra commits came in. I'll be happy to put it back on "main" if that's the better option. I'll also address the performance review remark.

Cheers!

The test data will be merged separately. Otherwise the build will fail for all implementations.

Issue #412, opened 2025-06-17
cpp: Acceptance tests do not pass on windows


musicinmybrain:
### 👓 What did you see?

In a [package review for a new `gherkin` package in Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=2369997), the reviewer noted that the C and C++ implementations have acceptance tests, that they don’t appear to be part of the CI configuration, and that they currently fail.

### ✅ What did you expect to see?

I expected that all acceptance tests would pass.

### 🔬 How could we reproduce it?

Working on Fedora 42:

```
$ git clone https://github.com/cucumber/gherkin.git
$ cd gherkin
$ make -C c acceptance
[…]
mkdir -p acceptance/testdata/good
echo bin/gherkin_generate_tokens
bin/gherkin_generate_tokens
bin/gherkin_generate_tokens ../testdata/good/prefixed-keywords.feature > acceptance/testdata/good/prefixed-keywords.feature.tokens
diff --strip-trailing-cr --unified ../testdata/good/prefixed-keywords.feature.tokens acceptance/testdata/good/prefixed-keywords.feature.tokens
--- ../testdata/good/prefixed-keywords.feature.tokens   2025-06-17 07:49:09.772177553 -0400
+++ acceptance/testdata/good/prefixed-keywords.feature.tokens   2025-06-17 07:54:16.135316509 -0400
@@ -4,9 +4,9 @@
 (4:1)Other:/  In this scenario the parser should prefer the longest keyword./
 (5:1)Other://
 (6:3)ScenarioLine:()Senaryo/Erasing agent memory/
-(7:5)StepLine:(Context)Sipoze ke /there is agent J/
+(7:5)StepLine:(Context)Sipoze /ke there is agent J/
 (8:5)StepLine:(Conjunction)Ak /there is agent K/
 (9:5)StepLine:(Action)Le /I erase agent K's memory/
-(10:5)StepLine:(Outcome)Le sa a /there should be agent J/
+(10:5)StepLine:(Unknown)Le sa a /there should be agent J/
 (11:5)StepLine:(Conjunction)Men /there should not be agent K/
 EOF
make: *** [Makefile:100: acceptance/testdata/good/prefixed-keywords.feature.tokens] Error 1
make: *** Deleting file 'acceptance/testdata/good/prefixed-keywords.feature.tokens'
make: Leaving directory '/home/ben/src/forks/gherkin/c'
$ make -C cpp acceptance
[…]
[...]

PR #414, opened 2025-06-17
cpp: Enable codegen

mpkorstanje:
### 🤔 What's changed?

* Adds cpp to the root make file so the parser is updated with `make generate`
* Regenerate cpp parser

https://github.com/cucumber/gherkin/issues/412

### 📋 Checklist:


mpkorstanje:
Looking at https://github.com/cucumber/gherkin/pull/413 the acceptance tests pass except for https://github.com/cucumber/gherkin/issues/400.

musicinmybrain:
The (first) failing acceptance test in this case is:

```
diff --unified ../testdata/good/prefixed-keywords.feature.tokens acceptance/testdata/good/prefixed-keywords.feature.tokens
--- ../testdata/good/prefixed-keywords.feature.tokens   2025-06-17 07:49:09.772177553 -0400
+++ acceptance/testdata/good/prefixed-keywords.feature.tokens   2025-06-17 11:32:48.135765735 -0400
@@ -7,6 +7,6 @@
 (7:5)StepLine:(Context)Sipoze ke /there is agent J/
 (8:5)StepLine:(Conjunction)Ak /there is agent K/
 (9:5)StepLine:(Action)Le /I erase agent K's memory/
-(10:5)StepLine:(Outcome)Le sa a /there should be agent J/
+(10:5)StepLine:(Action)Le /sa a there should be agent J/
 (11:5)StepLine:(Conjunction)Men /there should not be agent K/
 EOF
```

which is exactly the example in https://github.com/cucumber/gherkin/pull/414.

mpkorstanje:
Thanks for the quick PR. We can fix this in small steps. I'll check/merge tomorrow.

To further fix things, in general terms, in the implementation there should be a `match_{somethint step related}` method that checks which of the keywords matches. It does this by checking the given/when/then keywords separately. These keywords should be combined and sorted on length before this is done.

The solutions to #400 might provide some inspiration.

But I'm on my phone right now so that's all I can usefully contribute for today.

PR #426, opened 2025-07-11
(dotnet) Migrate gherkin messages to cucumber messages

clrudolphi:
### 🤔 What's changed?

This PR enhances the .Net implementation such that the Message events and pickles that are generated use the `Cucumber/Messages` types instead of the `Gherkin.Cucumber.Messages.Types` which had been embedded in the Gherkin project.

Parity with other implementation languages.
This will also simplify the Reqnroll implementation of Messages.

### 📋 Checklist:

----


clrudolphi:
Tests run locally are successful. I'm not sure, from reading the build log, why the test-dotnet job failed. Can anyone help me out?

mpkorstanje:
>  I'm not sure, from reading the build log, why the test-dotnet job failed. Can anyone help me out?

```
bin/gherkin --no-source --no-pickles ../testdata/good/i18n_fr.feature | jq --sort-keys --compact-output "." > acceptance/testdata/good/i18n_fr.feature.ast.ndjson
```

Produced an AST with  `"keywordType": "CONTEXT"` instead of the expected `keywordType": "Context"`.

Is there perhaps a conflict between https://github.com/cucumber/gherkin/blob/main/dotnet/Gherkin/StepKeywordType.cs and https://github.com/cucumber/messages/blob/main/dotnet/Cucumber.Messages/generated/StepKeywordType.cs?

Why this only happens in the french feature files is indeed a bit puzzling.

mpkorstanje:
> Why this only happens in the french feature files is indeed a bit puzzling.

Without looking further into the details, I think that might be another bug actually. Something along the lines of https://github.com/cucumber/gherkin/issues/400. 

I think that if you try to replace the `StepKeywordType` in Gherkin with the one from messages you'll find your issue.

mpkorstanje:
On inspection this looks suspicious:

https://github.com/cucumber/gherkin/blob/main/dotnet/Gherkin/GherkinDialect.cs#L40-L49

By using `.ToDictionary(item => item.Key, item => item.First());` it assumes that the `AsteriskKeyword` is the only duplicate keyword. Which probably isn't true. This should be something like `item => item.Size() > 1 : Unknown : item.First();` 

Again apologies for any vagueness.

dotnet/Gherkin.Specs/Helper/CucumberMessageEnumConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+        protected internal CucumberMessageEnumConverter()
+        {
+            var type = typeof(T);
+            foreach (var field in type.GetFields(BindingFlags.Public | BindingFlags.Static))
+            {
+#pragma warning disable CS8605 // Unboxing a possibly null value.
+                var value = (T)field.GetValue(null);
+#pragma warning restore CS8605 // Unboxing a possibly null value.
+                var attribute = field.GetCustomAttribute<DescriptionAttribute>();
+                var name = attribute?.Description ?? field.Name;
```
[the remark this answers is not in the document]
clrudolphi: These enums are within the Messages package; and this converter code is borrowed from the test suite of the Messages package. The test for a missing enum description would be better placed there. I'll add it there later.

dotnet/Gherkin/CucumberMessages/AstMessagesConverter.cs
```diff
@@ -208,34 +192,43 @@ private Step ConvertStep(Ast.Step step)
...
+            step.Keyword,
+            ConvertStepKeywordType(step.KeywordType),
+            step.Text,
+            docString,
+            dataTable,
+            idGenerator.GetNewId()
+        );
+    }
+
+    private Io.Cucumber.Messages.Types.StepKeywordType ConvertStepKeywordType(StepKeywordType keywordType)
```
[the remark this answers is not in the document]
clrudolphi: That would require a breaking change to the Parser and its output. I was trying to avoid that. Many projects use the parser and the `AST` types it creates, but very few, if any so far, use the `AstMessagesConverter` to create Messages. (simply trying to minimize the blast radius)

clrudolphi:
> > Why this only happens in the french feature files is indeed a bit puzzling.
> 
> Without looking further into the details, I think that might be another bug actually. Something along the lines of #400.
> 
> I think that if you try to replace the `StepKeywordType` in Gherkin with the one from messages you'll find your issue.
> 
> > Tests run locally are successful
> 
> And this is probably because some lookup is stored in a map or set with non-determistic ordering.

Actually the cause is simple, but not sure of the path forward.
This fails on the french feature file b/c this is the *_first_* feature file in the test script for which the test suite is attempting to compare the generated `GherkinDocument`. All prior tests in the suite are comparing the tokens. Therefore, this is the first diff test that is looking at the output that includes Messages types (and enum values). The acceptance test files use the old `StepKeywordType` enumeration, but the revised code uses the `StepKeywordType` enum from Messages (which accounts for the difference in casing in the values). The test run stops after the first failure, otherwise we would have seen the other ast, source and pickle acceptance tests also failing for the same reason.

My doubt however is why this hasn't surfaced for the other implementation languages that have already adopted Messages. Their tests too should show a difference between what is output and what the acceptance files have as content.

clrudolphi:
I had not tried previously to regenerate the .net parser, so I attempted that by following the directions on the Contributing.MD page (spinning up a docker container and running the make commands).
That is failing b/c the dockerfile was auto-updated last week to use .net9 while berp needs .net8.

Was that a mistake?

mpkorstanje:
> Therefore, this is the first diff test that is looking at the output that includes Messages types (and enum values). The acceptance test files use the old StepKeywordType enumeration, but the revised code uses the StepKeywordType enum from Messages (which accounts for the difference in casing in the value

This PR writes the uppercased enum names. The expected value is the lower cased description on the enum constant. So presumably the serialisation isn't done correctly. The other implementations do seem to do this correctly.

https://github.com/cucumber/messages/blob/ad48503f9ff91d5e249928c833a90aa4326ed481/dotnet/Cucumber.Messages/generated/StepKeywordType.cs#L18-L19

>  That is failing b/c the dockerfile was auto-updated last week to use .net9 while berp needs .net8.

Did you try to downgrade that again?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org).
 This document is formatted according to the principles of [Keep A CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+- [.NET] Migrated to the use of Cucumber/Messages. Eliminated the built-in Gherkin.CucumberMessages.Types ([#426](https://github.com/cucumber/gherkin/pull/426))
```
mpkorstanje, line 10:
```suggestion
### Changed
- [.NET] Migrated to the use of Cucumber/Messages. Eliminated the built-in Gherkin.CucumberMessages.Types ([#426](https://github.com/cucumber/gherkin/pull/426))
```

dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
```diff
@@ -168,27 +168,24 @@ protected virtual void CompileScenarioOutline(List<Pickle> pickles,
...
-                       lastStepKeywordType == StepKeywordType.Unknown
-                    ? StepKeywordType.Context
-                    : lastStepKeywordType;
-            default:
-                return StepKeywordType.Unspecified;
-        }
+            Io.Cucumber.Messages.Types.StepKeywordType.CONTEXT => PickleStepType.CONTEXT,
+            Io.Cucumber.Messages.Types.StepKeywordType.ACTION => PickleStepType.ACTION,
+            Io.Cucumber.Messages.Types.StepKeywordType.OUTCOME => PickleStepType.OUTCOME,
+            Io.Cucumber.Messages.Types.StepKeywordType.UNKNOWN => PickleStepType.UNKNOWN,
```
mpkorstanje, line 182: Previously this would return `stepKeywordType`.

dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
```diff
@@ -168,27 +168,24 @@ protected virtual void CompileScenarioOutline(List<Pickle> pickles,
...
+            Io.Cucumber.Messages.Types.StepKeywordType.CONJUNCTION =>
+                (lastStepKeywordType == PickleStepType.UNKNOWN
+                                ? PickleStepType.CONTEXT
```
mpkorstanje, line 185: Note to self: This switch appears to be correctly migrated. But also different from (at least) the java implementation. 

A scenario with the `*` keyword would be all `Context` steps in .Net, but `Unknown` in Java.

```
 * Step A
 * Step b
 * Step c
```

dotnet/Gherkin/Gherkin.csproj
```diff
@@ -44,4 +44,8 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Cucumber.Messages" Version="28.*" />
```
mpkorstanje, line 48:
```suggestion
    <PackageReference Include="Cucumber.Messages" Version="[27.0, 29.0)"/>
```

That way Renovate can automatically upgrade the upper bound.

dotnet/Gherkin/CucumberMessages/Pickles/PickleCompiler.cs
mpkorstanje: @clrudolphi this diverges from other implementations. Can be fixed after this is merged. One thing at a time. :wink: 

https://github.com/cucumber/gherkin/pull/427

clrudolphi:
> > I'm not sure, from reading the build log, why the test-dotnet job failed. Can anyone help me out?
> 
> ```
> bin/gherkin --no-source --no-pickles ../testdata/good/i18n_fr.feature | jq --sort-keys --compact-output "." > acceptance/testdata/good/i18n_fr.feature.ast.ndjson
> ```
> 
> Produced an AST with `"keywordType": "CONTEXT"` instead of the expected `keywordType": "Context"`.
> 

I found the problem within the CLI section of the testing project. It was using a default enum converter. I've changed that to use the proper one for the Messages enums. All tests pass now.

mpkorstanje (approved):
Looks good to me. But I assume you'll want to fix the `StepKeywordType`?

clrudolphi:
> On inspection this looks suspicious:
> 
> https://github.com/cucumber/gherkin/blob/c06186fa123a30cbbc273cc3894d7134038994b4/dotnet/Gherkin/GherkinDialect.cs#L40-L49
> 
> By using `.ToDictionary(item => item.Key, item => item.First());` it assumes that the `AsteriskKeyword` is the only duplicate keyword. Which probably isn't true. This should be something like `item => item.Size() == 1 ? Unknown : item.First();`
> 
> Again apologies for any vagueness. I can't actually run this code locally.

I haven't looked into this, but consider it out of scope for this PR. This is within the parser and not anything I'm touching right now.

mpkorstanje:
> I haven't looked into this, but consider it out of scope for this PR. This is within the parser and not anything I'm touching right now.

No problem.

https://github.com/cucumber/gherkin/pull/435

PR #436, opened 2025-07-19
Improving parser performance

jkronegg:
### 🤔 What's changed?

Several small performance improvements that make the parser about 18% faster.

Making Cucumber the greenest BDD test framework by having high performance engine.

The 18% performance improvement have been tested on the following JMH benchmark:

```java
import io.cucumber.messages.IdGenerator;
import io.cucumber.messages.types.Envelope;

import java.io.IOException;
import java.nio.file.Paths;
import java.util.stream.Stream;

public class MyClassBenchmark {
/*
very_long.feature
MyClassBenchmark.performance_optimisations2  thrpt   25  1444.817 ± 72.380  ops/s = +18%
MyClassBenchmark.main                        thrpt   25  1220.832 ± 67.814  ops/s

 */
    @Benchmark
    public Stream<Envelope> myMethod0() throws IOException {
        GherkinParser gherkinParser = GherkinParser.builder()
                .idGenerator(new IdGenerator() {
                    int id = 0;

                    @Override
                    public String newId() {
                        return Integer.toString(id++);
                    }
                })
                .build();
        return gherkinParser.parse(Paths.get("../testdata/good/very_long.feature"));
    }

}
```

I tracked down every millisecond, and found several improvements, so that in the end, small streams make big rivers. But that means that sometimes we change the code to gain just a few milliseconds, so not everyone may like it and take the risk to change the code.

Thus, each commit in this PR contains a given performance improvement. Each commit has been designed as a individual commit so could theoretically be cherry picked. That mean that the unwanted optimizations can be avoided easily.

Some example of optimization techniques used:
- avoid dynamic lambda creation
- loop unrolling
- precomputing String length
- replacing String instanciation by String char-by-char examination
- specific treatment depending on use-case (one-letter separator, empty collection, pre-sized collection, ...)
- use of specific interface to pass parameters (`Supplier` -> `BiConsumer`/`BiPredicate`)
- replaced `for-each` by classical `for`

I suggest to review the code commit by commit, starting sequentially from the first commit.

Note that after this PR, the only way I see to improve the performance is to use Java 11+ features (`String.lines()`, reusable immutable collections). I expect another 10% performance improvement when switching to Java 11+ features.

### 📋 Checklist:

- [x] I've changed the behaviour of the code (`Locations`)


mpkorstanje:
Cheers! Appreciate the commit structure. I'll have a good look on Monday.

> @jkronegg jkronegg requested a review from Copilot

I don't think we'll see those energy savings realized now. 😉

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
```diff
@@ -44,27 +52,24 @@ public void reset() {
...
+        token.matchedGherkinDialect = currentDialect;
+        token.matchedIndent = indent;
         token.location = atColumn(token.location, token.matchedIndent + COLUMN_OFFSET);
     }
 
     @Override
     public boolean match_EOF(Token token) {
         if (token.isEOF()) {
-            setTokenMatched(token, TokenType.EOF, null, null, null, null, null);
+            setTokenMatched(token, TokenType.EOF, null, null, 0, null, null);
```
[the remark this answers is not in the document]
jkronegg: As this is a private method, we can accept 7 parameters because the maintainability issue is limited to the class. Having a parameter object will just move the maintainability issue to another part of the code. Both the builder pattern an the parameter object will lead to object instanciation, which will have a negative impact on performance.

java/src/main/java/io/cucumber/gherkin/StringUtils.java
```diff
@@ -114,4 +115,21 @@ private static boolean contains(char[] characters, char c) {
         return false;
     }
 
+    public static class TrimmedText {
```
mpkorstanje, line 118: As this represents a text with indentation, it would be more appropriately called indented text.

java/src/main/java/io/cucumber/gherkin/StringUtils.java
```diff
@@ -114,4 +115,21 @@ private static boolean contains(char[] characters, char c) {
...
+        private final String text;
+        private final int indent;
+
+        public TrimmedText(String text, int indent) {
```
mpkorstanje, line 122: Complete nitpick. But can you flip these arguments? That way from left to right reads  "indent - text".

java/src/main/java/io/cucumber/gherkin/StringUtils.java
```diff
@@ -38,25 +36,28 @@ static String rtrim(String s) {
...
         int start = findFirstIndexNotIn(input, input.length(), whitespaceChars);
         int end = findLastIndexNotIn(input, start, whitespaceChars);
 
         String trimmed = input.substring(start, end);
         int indent = input.codePointCount(0, start);
-        return new SimpleEntry<>(trimmed, indent);
+        return new TrimmedText(trimmed, indent);
+        // the object instance is not truly created because
+        // the code is inlined by the hotspot compiler
+        // (as "-XX:+EliminateAllocations" is enabled by default).
```
mpkorstanje, line 60: Comments should be above the line they relate to.

Also this is quite cool!

mpkorstanje (commented):
Reviewed f74ceb45357caedee0eadc2b58472e27e8bfb994.

mpkorstanje (commented):
Reviewed 37a127565b993d5d2f93f24055a3499f65f2289b. 

No comments on this commit. 

But with JUnit 6  being scheduled for September, I think it is safe to start the optimizations for Java 17. Not in this PR, but as a separate set of PR's focused only on optimizations that are possible with Java 17.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -78,26 +77,52 @@ List<GherkinLineSpan> parseTags() {
         if (empty) {
             return emptyList();
         }
-        String uncommentedLine = StringUtils.removeComments(text);
-        int indexInUncommentedLine = 0;
 
-        String[] elements = uncommentedLine.split(TAG_PREFIX);
-        if (elements.length == 0) {
+        // look for the first tag
+        int indexStartTag = text.indexOf(TAG_PREFIX_CHAR);
```
mpkorstanje, line 92: Because there is a next tag too, please call this `indexStartCurrentTag`. This signals in advance that there will be pointers to other tags too.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -78,26 +77,52 @@ List<GherkinLineSpan> parseTags() {
...
+                indexEndTag--;
             }
-            tags.add(new GherkinLineSpan(column, TAG_PREFIX + token));
-            indexInUncommentedLine += element.length() + 1;
+            indexEndTag++;
+
+            if (indexEndTag > indexStartTag + 1) {
+                // non-empty tag found
+
+                // check that the tag does not contain whitespace characters
```
mpkorstanje, line 119: This comment should sit on top of the for loop

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -78,26 +77,52 @@ List<GherkinLineSpan> parseTags() {
...
+                int symbolLength = text.codePointCount(0, indexStartTag);
+                int column = indent + symbolLength + COLUMN_OFFSET;
+                for (int i = indexStartTag + 1; i < indexEndTag; i++) {
+                    if (text.charAt(i) <= ' ') {
```
mpkorstanje, line 123: Functionally this isn't equivalent to `containsWhiteSpace`.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -78,26 +77,52 @@ List<GherkinLineSpan> parseTags() {
...
+                        // skip whitespace characters
```
mpkorstanje, line 124: This comment makes no sense.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -78,26 +77,52 @@ List<GherkinLineSpan> parseTags() {
...
+                indexNextStartTag++;
             }
-            int symbolLength = uncommentedLine.codePointCount(0, indexInUncommentedLine);
-            int column = indent + symbolLength + COLUMN_OFFSET;
-            if (containsWhiteSpace(token)) {
-                throw new ParserException("A tag may not contain whitespace", Locations.atColumn(location, column));
+
+            // look for the end of current tag (going back from begin of next tag)
+            indexEndTag = indexNextStartTag - 1;
+            while (indexEndTag > indexStartTag && text.charAt(indexEndTag) <= ' ') {
```
mpkorstanje, line 111: `text.charAt(indexEndTag) <= ' '` looks suspicious here. Previously was done by `rtrim` which trimmed on `WHITESPACE_CHARS_EXTENDED` .

mpkorstanje (commented):
Reviewed 285da13cf410bbfd656061bb0fe4e1e3bedf545d.

My biggest objection is the simplified whitespace handling. If we can keep that equivalent to the current implementation I'm happy to see this continue.

Then on a separate PR I'd be happy to discuss the possibility of using `Character.isWhitespace` to replace the current whitespace detection all together and update the test data accordingly.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -69,7 +71,11 @@ boolean startsWith(String prefix) {
     }
 
     String substringTrimmed(int beginIndex) {
-        return text.substring(beginIndex).trim();
+        // trim the beginning of the line (the end of line has already been trimmed in the constructor)
+        while ((beginIndex < textLength) && (text.charAt(beginIndex)<=' ')) {
```
mpkorstanje: Nice optimization.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -70,35 +70,69 @@ boolean startsWith(String prefix) {
...
             return emptyList();
         }
-        List<GherkinLineSpan> tags = new ArrayList<>(elements.length);
-        for (String element : elements) {
-            String token = rtrim(element);
-            if (token.isEmpty()) {
-                continue;
+
+        List<GherkinLineSpan> tags = new ArrayList<>();
+        int indexEndOfLine = indexComment >= 0 ? indexComment : text.length();
```
jkronegg, line 100: Oops, I didn't know that a tag can contain a '#' (e.g. "@comment_tag#2" like in `tags.feature`). This is why the "make acceptance" fails. I'll provide a correction.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -179,10 +185,9 @@ List<GherkinLineSpan> parseTableCells() {
     }
 
     boolean startsWithTitleKeyword(String keyword) {
-        int keywordLength = keyword.length();
```
mpkorstanje, line 158: This value is used twice. Doesn't in-lining it hurt performance?

java/src/test/java/io/cucumber/gherkin/GherkinLineTest.java
```diff
@@ -218,6 +218,24 @@ void startsWithTitleKeyword() {
         assertFalse(gherkinLine.startsWithTitleKeyword("Feat"));
     }
 
+    @Test
+    void startsWithTitleKeyword_short() {
```
mpkorstanje, line 222: `_short` and `_mini` could be more descriptive of the condition that is actually being tested.

mpkorstanje (commented):
Reviewed e54e9c58e0e6dae502223dcfe3675eec514bf9d4

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -70,6 +70,10 @@ boolean startsWith(String prefix) {
         return text.startsWith(prefix);
     }
 
+    public boolean startsWith(char tableCellSeparator) {
```
mpkorstanje, line 73: Rename `tableCellSeparator` to `prefix` to match the other `startsWith` method.

mpkorstanje (commented):
Reviewed 63a760b28ae439ce226eb1924db2d31a2b45b608.

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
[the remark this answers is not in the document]
mpkorstanje: We don't have to discuss this with the AI. I'm disregarding all it's comments. This one also sounds like it ran checkstyle and is now repeating the output of checkstyle back to us.

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
```diff
@@ -46,25 +46,21 @@ public void reset() {
...
-
-    private void setTokenMatched(Token token, TokenType matchedType, String text, String keyword, Integer indent, StepKeywordType keywordType, List<GherkinLineSpan> items) {
+    private void setTokenMatched(Token token, TokenType matchedType, String text, String keyword, int indent, StepKeywordType keywordType, List<GherkinLineSpan> items) {
         token.matchedType = matchedType;
         token.matchedKeyword = keyword;
         token.keywordType = keywordType;
         token.matchedText = text;
         token.matchedItems = items;
-        token.matchedGherkinDialect = getCurrentDialect();
-        token.matchedIndent = indent != null ? indent : (token.isEOF() ? 0 : token.line.getIndent());
```
mpkorstanje, line 60: I'm assuming you used an automated refactoring tool to do this? If so, ok. If not, please revert and redo with a tool.

java/src/test/java/io/cucumber/gherkin/GherkinDialectsTest.java
```diff
@@ -0,0 +1,17 @@
+package io.cucumber.gherkin;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class GherkinDialectsTest {
+    @Test
+    void each_dialect_has_at_least_one_given_keyword_except_the_star() {
```
mpkorstanje, line 9: Nice find. If we test this, this should be a language agnostic test against `gherkin-language.json`. But I'd rather fix this in Cucumber-JVM with a "Generic" `@Step` annotation/method to fall back to if we can't work out the previous GWT. 

Could you please create an issue for that against Cucumber-JVM?

mpkorstanje (commented):
Reviewed 636d38fb13891f4f55f38f8987ecc1c9e0ef2a25

Please revert.

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
jkronegg: This is hand-made refactoring without a tool. I understand that you trust a tool more than a developer for such refactoring, but here the replacement is quite simple: when the `indent` argument is passed with `null` value, replace by `token.line.getIndent()` (except in `match_EOF(Token)` where we replace by 0).

I agree that tool-based refactoring are useful, typically for large-scale refactoring. But here it was really simple, and I don't know which refactoring tool to use.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
jkronegg: I saw comments that start at the beginning of the line (so without whitespace before the '#'), e.g. in `descriptions_with_comments.feature:16`:

    Scenario: without indentation
    This is a description without indentation
    # comment
    and a comment in the middle and at the end
    # comment 2

Is the `<WHITESPACE_CHARS>#` only for lignes that contains tags ?
Or is the true comment expression `^#|<WHITESPACE_CHARS>#` ?

mpkorstanje: No, only for lines with a tag. I.e. lines that start with `@` after an left trim.

mpkorstanje: Lines with a tag are also the only lines that allow trailing comments. On every other line it would be part of the step, title, description, ect.

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
mpkorstanje: I was suspicious because this method has 10 calls sites, and only 8 were updated. Now I've done the refactoring with IDEA (Eclipse would work too) and it looks all correct. The two remaining call sites match `Other` and `Comment` tokens but those use a constant indent of 0 

The recipes I used

1. Extract `indent != null ? indent : (token.isEOF() ? 0 : token.line.getIndent());` to a local variable at the top of the `setTokenMatched` method.
2. Extract `setTokenMatched2` from `setTokenMatched` excluding that new line. 
3. Inline `indent != null ? indent : (token.isEOF() ? 0 : token.line.getIndent());` again.
4. Inline `setTokenMatched`
5. Rename `setTokenMatched2` to `setTokenMatched`
6. Resolve all constant conditions at `setTokenMatched` invocations.

java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
```diff
@@ -173,6 +185,15 @@ public Set<StepKeywordType> getStepKeywordTypesSet(String keyword) {
         return stepKeywordTypes;
     }
 
+    public StepKeywordType getStepKeywordType(String keyword) {
```
mpkorstanje, line 189: No need to add this method to the public API.

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
```diff
@@ -180,26 +189,71 @@ private boolean match_DocStringSeparator(Token token, String separator, boolean
         return false;
     }
 
+    private void prepareStepLineLoopUnrolling() {
```
mpkorstanje, line 196: The `currentDialect` is mutable. I don't see how this would work without some sort of cache invalidation.

mpkorstanje (commented):
Reviewed 034bd70b66d9e57e12cd58060ea6c665077c12f4.

I think this is more complicated than it seems. For the benefit of getting all the less complicated stuff merged, please move to a separate PR.

java/src/main/java/io/cucumber/gherkin/Locations.java
```diff
@@ -46,17 +44,13 @@ private static Long getLong(int i) {
     }
 
     static Location atColumn(Location location, int column) {
-        requireNonNull(location);
-        if (column <= 0) {
-            throw new IllegalArgumentException("Columns are index-1 based");
-        }
+        // By design, the location cannot be null (it comes from the Token)
+        // By design, the column cannot be lower than 0
```
mpkorstanje, line 48: Lower than 1 actually.

java/src/main/java/io/cucumber/gherkin/Locations.java
```diff
@@ -46,17 +44,13 @@ private static Long getLong(int i) {
...
         return new Location(location.getLine(), getLong(column));
     }
 
     static Location atLine(int line) {
-        if (line < 0) {
-            throw new IllegalArgumentException("Lines are index-0 based");
-        }
+        // By design, the line is never lower than 0 (it comes from TokenScanner)
```
mpkorstanje, line 53: Lower than 1 actually. The < 0 is a mistake.

mpkorstanje (commented):
Reviewed 99957cdf2d10f063a5ea4afb5a1040c6cab362ec

mpkorstanje (commented):
Reviewed e896df65e4ae450fc323d2deea363c3907f5892d.

Nice!

mpkorstanje (commented):
Reviewed 33f0c4557b1cfa66bc68767f4373d8fd2806a5ce :ok:

java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
```diff
@@ -98,25 +98,34 @@ private void compileRule(List<Pickle> pickles, Rule rule, List<Tag> parentTags,
     }
 
     private void compileScenario(List<Pickle> pickles, Scenario scenario, List<Tag> parentTags, List<Step> backgroundSteps, String language, String uri) {
-        List<PickleStep> steps = new ArrayList<>();
+        int pickleStepCount = scenario.getSteps().size() + backgroundSteps.size();
+        List<PickleStep> steps = new ArrayList<>(pickleStepCount);
         if (!scenario.getSteps().isEmpty()) {
-            List<Step> allSteps = new ArrayList<>();
+            // usually, a scenario has at least one step, but that's not mandatory
```
mpkorstanje, line 104: Comment should sit on the if.

java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
```diff
@@ -98,25 +98,34 @@ private void compileRule(List<Pickle> pickles, Rule rule, List<Tag> parentTags,
...
             }
         }
 
-        List<Tag> scenarioTags = new ArrayList<>();
-        scenarioTags.addAll(parentTags);
-        scenarioTags.addAll(scenario.getTags());
+        List<Tag> scenarioTags;
+        if (parentTags.isEmpty() && scenario.getTags().isEmpty()) {
+            // in most cases, the scenario will have no tag, so we avoid creating a new list
+            scenarioTags = emptyList();
```
mpkorstanje, line 123: You could  `scenarioTags = parentTags` and remove the `parentTags.isEmpty()` check.

java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
```diff
@@ -98,25 +98,34 @@ private void compileRule(List<Pickle> pickles, Rule rule, List<Tag> parentTags,
     }
 
     private void compileScenario(List<Pickle> pickles, Scenario scenario, List<Tag> parentTags, List<Step> backgroundSteps, String language, String uri) {
-        List<PickleStep> steps = new ArrayList<>();
+        int pickleStepCount = scenario.getSteps().size() + backgroundSteps.size();
+        List<PickleStep> steps = new ArrayList<>(pickleStepCount);
         if (!scenario.getSteps().isEmpty()) {
-            List<Step> allSteps = new ArrayList<>();
+            // usually, a scenario has at least one step, but that's not mandatory
+            List<Step> allSteps = new ArrayList<>(pickleStepCount);
```
mpkorstanje, line 105: I think we can avoid this array creation by duplicating the `allSteps` for loop.

mpkorstanje (commented):
Reviewed fcd53a50d63107499429b7bb75f482ade6e1cf28

java/src/test/java/io/cucumber/gherkin/GherkinDocumentBuilderTest.java
```diff
@@ -52,7 +52,16 @@ void parses_rules() {
...
+                "      Given b\n" +
+                "\n" +
+                "  @someTag\n" +
+                "  Scenario: with examples\n" +
+                "      Given the <value> minimalism\n" +
+                "\n" +
+                "      Examples:\n" +
+                "      | value |\n" +
+                "      | 1     |\n" +
+                "      | 2     |";
```
mpkorstanje, line 64: I don't understand what motivated this change. This test now looks unfocused. It now tests rules in combination with scenario outlines and tags and backgrounds. But only to a limited degree.

And I would expect these to (already) be part of the acceptance test suite.

mpkorstanje (commented):
Reviewed b6607c9a2e3ce9f34644e7bddaacc2565933a8d7

mpkorstanje (commented):
Reviewed ffa3fe4550bb19cd350d56c65090275cd3be6ac2

mpkorstanje (commented):
Reviewed ff49370f559fcf0967bf49ed315c14564626af7a

java/src/main/java/io/cucumber/gherkin/GherkinDocumentBuilder.java
```diff
@@ -245,32 +245,41 @@ private static String joinMatchedText(List<Token> lineTokens, int endIndex) {
         return content.toString();
     }
 
+    @SuppressWarnings("ForLoopReplaceableByForEach") // classic 'for' loop is ~2x faster than 'for-each'
     private List<TableRow> getTableRows(AstNode node) {
         List<Token> tokens = node.getTokens(TokenType.TableRow);
-        List<TableRow> rows = new ArrayList<>(tokens.size());
+        int tokenCount = tokens.size();
```
mpkorstanje, line 251: Nitpick: 
 * count -> a number obtained by iterating over some array or list
 * size ->  a property of a collection
 * length -> a property of an array.
 
 So I think using `tokenSize` here would be the correct choice.

mpkorstanje (commented):
Reviewed c4c33c17299b302bec0292b146079a4aef5f87ad

java/src/test/java/io/cucumber/gherkin/GherkinParserTest.java
```diff
@@ -42,10 +45,43 @@ void use_this_in_readme() {
     @Test
     void test_for_profiler_parser() throws IOException {
         for (int i = 0; i < 1000; i++) {
-            GherkinParser.builder().build().parse(Paths.get("../testdata/good/very_long.feature"));
+            GherkinParser.builder()
```
mpkorstanje, line 48: Kinda surprised this isn't extracted out of the loop. I would not expect that reusing the parser would influence the profiler.

mpkorstanje (commented):
Reviewed 597607a3b4fa8c1a8dd07f26d95b34fff343a5f5

mpkorstanje (changes requested):
Overall impression: This looks good!

There are some sketchy bits around whitespace handling. Gherkin uses an extended whitespace definition to catch problems with accidentally inserted non-standard whitespace characters. I.e. when someone copies a feature from MS Office, Outlook or some other rich text editor. Unfortunately, the test coverage for that is quite limited, so I'm not confident we aren't breaking things. But I think most of that is fixable by retaining the old behavior more precisely.

I would like to put the loop unrolling (https://github.com/cucumber/gherkin/commit/034bd70b66d9e57e12cd58060ea6c665077c12f4) on a separate PR. The loop unrolling introduces state that should be nicely isolated and isn't quite trivial.

Given that `test_for_profiler_parser_with_summary_statistics` also produces a 400ms difference between the fastest and the slowest. I'm also not quite convinced about the necessity. I would expect Java to execute these instructions in parallel with or without unrolling. If not, then eventually hotspot should catch on and unroll the loops right?

And please revert https://github.com/cucumber/gherkin/commit/636d38fb13891f4f55f38f8987ecc1c9e0ef2a25

Beyond that I've only got minor comments and nitpicks. If we focus on getting those done then we can merge this PR quickly. If we focus on the unrolling in a separate PR (possibly concurrently with this one) then that won't have to block the easy changes.

java/src/main/java/io/cucumber/gherkin/GherkinDocumentBuilder.java
```diff
@@ -245,32 +245,41 @@ private static String joinMatchedText(List<Token> lineTokens) {
...
-        for (TableRow row : rows) {
+        for (int i = 0, rowsSize = rows.size(); i < rowsSize; i++) {
+            TableRow row = rows.get(i);
             if (row.getCells().size() != cellCount) {
                 throw new ParserException.AstBuilderException("inconsistent cell count within the table", row.getLocation());
             }
         }
     }
 
+    @SuppressWarnings("ForLoopReplaceableByForEach") // classic 'for' loop is ~2x faster than 'for-each'
```
mpkorstanje, line 276: How was this established?  I can't seem to reproduce it.

I've done a very small sample test. 3 samples with and without the optimization. With the optimization enabled I get between 330ms and 380ms for `getCells`. Without the optimization 280ms to 480ms. Variability aside, it's not a 2x difference.

java/src/test/java/io/cucumber/gherkin/GherkinParserTest.java
```diff
@@ -42,10 +45,43 @@ void use_this_in_readme() {
...
+                            return Integer.toString(id++);
+                        }
+                    })
+                    .build().parse(Paths.get("../testdata/good/very_long.feature"));
         }
     }
 
+    @Test
+    @Disabled
+    void test_for_profiler_parser_with_summary_statistics() throws IOException {
```
mpkorstanje, line 62: Both of these could be replaced with https://github.com/cucumber/gherkin/pull/437.

mpkorstanje:
@jkronegg I'm happy with the current performance improvements. If you agree, then I think we can work towards merging.

TODO:

1. Check for dead code
  * StringUtils.removeComments
  * StringUtils.rtrim
  
2. Make sure we have unit test coverage 

Possibilities for further improvements on the next PR:

1. The `WHITESPACE_CHARS` characters can be checked with a single bit mask. See `Character.isSpace` for an example.
2. The `WHITESPACE_CHARS_EXTENDED` aren't but that's only two extra checks.
3. The `WHITESPACE_CHARS_EXTENDED_KEEP_NEW_LINES` idem.

java/src/main/java/io/cucumber/gherkin/GherkinDocumentBuilder.java
jkronegg: The 2x factor was determined using the profile, not through micro-benchmark. With JMH, I would expect to have a lower difference, or no difference at all if the JIT compiler works well. The results when reverting the commit is unexpected (for-each being faster than classical for). That's a behavior that is noticed in https://reneschwietzke.de/java/java-loop-performance.html but only when disabling the JIT compiler.

As we have no evidence of better performance with the classical for, so let's keep the for-each, this makes the code more readable.

java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
```diff
@@ -126,6 +116,17 @@ private void compileScenario(List<Pickle> pickles, Scenario scenario, List<Tag>
         pickles.add(pickle);
     }
 
+    private List<Tag> compileTags(List<Tag> parentTags, List<Tag> scenarioTags) {
+        if (scenarioTags.isEmpty()) {
+            // in most cases, the scenario will have no tag, so we avoid creating a new list
+            return parentTags;
```
jkronegg, line 124: Nice optimization

java/src/main/java/io/cucumber/gherkin/StringUtils.java
```diff
@@ -88,13 +88,17 @@ private static int findFirstIndexNotIn(String input, int endIndex, char[] charac
 
     static boolean containsWhiteSpace(String input, int fromIndex, int toIndex) {
         for (int i = fromIndex; i < toIndex; i++) {
-            if (contains(WHITESPACE_CHARS, input.charAt(i))) {
+            if (contains(WHITESPACE_CHARS_EXTENDED, input.charAt(i))) {
```
jkronegg, line 94: Could be replaced by `if (isWhiteSpace(input.charAt(i))) {`.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -139,11 +163,11 @@ List<GherkinLineSpan> parseTableCells() {
                         // Skip the first empty span
                         beforeFirst = false;
                     } else {
-                        Entry<String, Integer> trimmedCellIndent = trimAndIndentKeepNewLines(cellBuilder.toString());
-                        int column = indent + cellStart + trimmedCellIndent.getValue() + COLUMN_OFFSET;
-                        lineSpans.add(new GherkinLineSpan(column, trimmedCellIndent.getKey()));
+                        StringUtils.IndentedText trimmedCellIndent = trimAndIndentKeepNewLines(cellBuilder.toString());
```
mpkorstanje, line 166: Same as `parseTags`, `parseTableCells` depends on a condition in the matcher. Extract out into `TableLineParser.parse(int indent, String text)`.

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
jkronegg: My mistake, it was done in another commit => I'll do a separate PR.

java/src/test/java/io/cucumber/gherkin/GherkinDocumentBuilderTest.java
jkronegg: I don't remember the rationale behind this addition, so I reverted to the original test case.

jkronegg:
> 1. Check for dead code
> 
> * StringUtils.removeComments
> * StringUtils.rtrim

I removed these two methods, but noticed that `GherkinLine.substringTrimmed` was not managing the spaces that are before and after the rule name consistently: 
- the spaces that are after the rule name were trimmed using `StringUtils.trimAndIndent` (so with `WHITESPACE_CHARS_EXTENDED`
- the spaces that are before the rule name were trimmed using `String.trim` so everything that `char<=' '` was removed (without support for extended whitespace chars).

I changed the behavior to still have the `String.trim` behavior (for backward compatibility) and added the support for extended whitespaces to have a coherent trimming method for spaces that are before and after the rule name.

java/src/main/java/io/cucumber/gherkin/GherkinLine.java
```diff
@@ -70,35 +71,60 @@ boolean startsWith(String prefix) {
...
+    public boolean startsWith(char prefix) {
+        return !empty && text.charAt(0) == prefix;
+    }
+
     String substringTrimmed(int beginIndex) {
-        return text.substring(beginIndex).trim();
+        // trim the beginning of the line (the end of line has already been trimmed in the constructor)
+        while ((beginIndex < textLength)) {
+            char c = text.charAt(beginIndex);
+            if (!(c <= ' ' || StringUtils.isWhiteSpaceExtended(c))) break;
```
mpkorstanje, line 82: On reflection, between https://github.com/cucumber/gherkin/pull/442 catching most of the space like characters `< ' '` and the other characters being line noise, I would be happy breaking this.

Please do extract this as `leftTrim` to StringUtils`.

java/src/main/java/io/cucumber/gherkin/StringUtils.java
```diff
@@ -104,6 +104,15 @@ static boolean containsWhitespace(String input, int fromIndex, int toIndex) {
         return false;
     }
 
+    static String substringAndTrim(String input, int beginIndex) {
```
jkronegg, line 107: The new method is not very efficient, as seen in the profiler. Checking whether the character is a space or not is about 50% of the method duration, and the `text` length pre-computed in `GherkinLine` constructor is not reused:
image
To be managed later...

mpkorstanje: Interesting. `StringUtils.isWhiteSpace` only takes 16% of `StringUtils.substringTrimmed` for me. 

image

>  and the text length pre-computed in GherkinLine constructor is not reused:

Ah good point. Didn't realize there would be call to `String.coder()` for every `length()` call.

mpkorstanje: https://github.com/cucumber/gherkin/pull/444

mpkorstanje:
And after #445 we have now realized the +24% improvement.

`main` @ dcbf4be73de56ec530af5239fec1a6a1c3048ff3

```
GherkinParserBenchmarkTest.parse  thrpt    5  1792.014 ± 10.810  ops/s
```

`main` @3a0b21a58b8baf30c4675c6c8710357d83dfaf40 

```
GherkinParserBenchmarkTest.parse  thrpt    5  2224.063 ± 15.719  ops/s
```

PR #443, opened 2025-07-27
StepLine identification loop unrolling

jkronegg:
### 🤔 What's changed?

In the `GherkinTokenMatcher`, the `match_StepLine(Token)` method has been optimized using a loop unrolling technique.

Parsing the BDD files faster.

When profiling `GherkinParserBenchmarkTest.profileParse()` on the main branch, we see that `match_StepLine(Token)` uses about half of the time to manage the loop and the keywords:
image

The new code consists to unroll the keywords loop for the first 6 step keywords of the language (each language has at least 6 keywords). The 6 keywords length and type is pre-computed.

After the loop unrolling, the profiler shows that loop iterator and `getStepKeywordType` impact is not visible anymore: 
image

In terms of performance the profiler shows for `GherkinTokenMatcher.match_StepLine(Token)`:
- main branch from release 33.1.0 : 419 ms
- this version with StepLine loop unrolling: 297 ms (~29% faster)

With the JMH benchmark, the effect is not visible, but on realworld project with 740 BDD scenarios, the ~30% improvement is also visible on the profiler flame graph.

### 📋 Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org).
 This document is formatted according to the principles of [Keep A CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Changed
+- [Java] Optimize GherkinTokenMatcher performance ([#443](https://github.com/cucumber/gherkin/pull/443))
 
 ## [33.1.0] - 2025-07-27
 ### Changed
--- a/java/src/main/java/io/cucumber/gherkin/GherkinLine.java
+++ b/java/src/main/java/io/cucumber/gherkin/GherkinLine.java
@@ -73,4 +73,7 @@ class GherkinLine {
                text.startsWith(keyword);
     }
 
+    public int getTextLength() {
+        return textLength;
+    }
 }
--- a/java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
+++ b/java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
@@ -33,6 +33,7 @@ class GherkinTokenMatcher implements TokenMatcher {
     private StepLineElem stepLineElem4;
     private StepLineElem stepLineElem5;
     private List<StepLineElem> remainingStepLineElems;
+    private String stepLineElemsLanguage;
 
     GherkinTokenMatcher(GherkinDialectProvider dialectProvider) {
         this.dialectProvider = dialectProvider;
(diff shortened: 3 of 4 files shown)


mpkorstanje (commented):
I can't come up with a way to either coax the JVM into unrolling this code on its own, or a data structure that is faster. So unrolling it is. The performance gains are just too good.

What I don't like however is all the state sitting around in `GherkinTokenMatcher`, the 6 keyword minimum requirement and the fact that the optimization doesn't work so well once there are more keywords. And this optimization is only applied for step keywords while the feature, scenario and scenario outline keywords all have the same problem.

However I think we can solve most of those problems with code generation.

```java
interface KeywordMatcher {

    String matchStepKeyword(GherkinLine line);
    String matchFeatureKeyword(GherkinLine line);
    ....
}
```

The `KeywordMatcher` would be an adjunct of the `GherkinDialect`, and each language would get its own generated implementation for matching.

```java
class <language> implements KeywordMatcher {
    private final String stepKeyword0;
    private final String stepKeyword1;
    ....
    private final String stepKeywordN;


    String match(GherkinLine line) {
        if (line.startsWith(stepKeyword0)) {
            return stepKeyword0;
        }

        if (line.startsWith(stepKeyword1)) {
            return stepKeyword1;
        }
        ....
        if (line.startsWith(stepKeywordN)) {
            return stepKeywordN;
        }
        return null;
    }
   ...
}
```

Before spending time on creating the code generation it would probably make sense to hand code the English version and use it just for step keywords to validate the performance improvements.

jkronegg:
> What I don't like however is all the state sitting around in `GherkinTokenMatcher`, the 6 keyword minimum requirement and the fact that the optimization doesn't work so well once there are more keywords.

Would it be acceptable that `GherkinDialect` provide the step keywords as a `List<StepLineElem>` ? This will allow `GherkinTokenMatcher` to use the pre-computed values, then maybe the JVM will do the loop unrolling automatically. The disadvantage is that we put parsing-related data structures in the `GherkinDialect`.

mpkorstanje:
On  main today:

```
Benchmark                          Mode  Cnt     Score   Error  Units
GherkinParserBenchmarkTest.parse  thrpt    5  2047.922 ± 7.949  ops/s
```

On `step_line_loop_unrolling`:

```
GherkinParserBenchmarkTest.parse  thrpt    5  2098.978 ± 20.299  ops/s
```

On `step_line_optimized_parsing`:

```
GherkinParserBenchmarkTest.parse  thrpt    5  2108.770 ± 18.839  ops/s
```


Starting fresh from main I've added the (to be) generated key word matcher on: https://github.com/cucumber/gherkin/tree/generated-matcher

```
GherkinParserBenchmarkTest.parse  thrpt    5  2149.722 ± 14.959  ops/s
```

This is already faster than both `step_line_loop_unrolling` and `step_line_optimized_parsing` and outside of the error bars too.


I've made the matcher return both keyword and keyword type on On https://github.com/cucumber/gherkin/tree/generated-matcher-with-keyword-type

```
GherkinParserBenchmarkTest.parse  thrpt    5  2142.330 ± 13.350  ops/s
```

But that didn't make much of a difference. 

Finally I added the length as well on:

https://github.com/cucumber/gherkin/tree/generated-matcher-with-keyword-type-and-length

```
GherkinParserBenchmarkTest.parse  thrpt    5  2166.756 ± 22.610  ops/s
```

Which does make a a difference though we're within reach of the error bars.

mpkorstanje:
@jkronegg could you compare my branches on your machine?

mpkorstanje:
On https://github.com/cucumber/gherkin/tree/generated-matcher-with-keyword-type-and-length I just got:

```
GherkinParserBenchmarkTest.parse  thrpt    5  2215.207 ± 13.806  ops/s
```

By making classes final!

mpkorstanje:
Superseded by https://github.com/cucumber/gherkin/pull/445, but that wasn't possible without identifying the potential for loop unrolling you did here.

jkronegg:
> Superseded by #445, but that wasn't possible without you identifying the potential for loop unrolling here.

You're welcome, it was a pleasure. I'm happy to have inspired the great solution with the generated matcher☺️.

PR #445, opened 2025-07-30
java: Generate keyword matchers

mpkorstanje:
### 🏷️ What kind of change is this?

- :book: Documentation (improvements without changing code)
- :bank: Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)
- :bug: Bug fix (non-breaking change which fixes a defect)
- :zap: New feature (non-breaking change which adds new behaviour)
- :boom: Breaking change (incompatible changes to the API)

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)
- [ ] I've changed the behaviour of the code
  - [ ] I have added/updated tests to cover my changes.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
- [ ] Users should know about my change
  - [ ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

PR #449, opened 2025-07-30
Minor performance improvement after keyword match unrolling

jkronegg:
### 🤔 What's changed?

After #445 , there is still some performance improvements:
- replaced `Line.startsWithTitleKeyword` by `Line.startsWith` in the generated matcher
- cached the matcher functions to avoid dynamic lambda creation

I also inverted some "if" conditions and renamed `KeywordMatcher.matchExampleKeyword` to `matchExamplesKeyword` (added missing "s").

Finish the unrolling story.

On main, we have:
image

The modified branch shows that we now use the `Line.startsWith` and we don't have the dynamic lambda call anymore:
image

There is no real String concatenation in the generated matcher because `"${feature.keyword}"` and ` TITLE_KEYWORD_SEPARATOR` are two compile-time constants that will be concatenated at compile-time.

### 📋 Checklist:


java/src/test/java/io/cucumber/gherkin/KeywordMatchersTest.java
```diff
@@ -34,6 +37,20 @@ void featureKeywordsAreConsistent(GherkinDialect dialect) {
         }
     }
 
+    @Test
+    void featureKeywordsDontMatchStringThatStartsSimilarlyButIsNotAFeatureKeyword() {
+        // Given
+        KeywordMatcher matcher = KeywordMatchers.of("en");
```
mpkorstanje, line 43: `Constants.DEFAULT_LANGAUGE`

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
```diff
@@ -90,71 +101,75 @@ public boolean match_Other(Token token) {
...
-        String language = matcher.group(1);
-        setTokenMatched(token, TokenType.Language, language, null, token.line.getIndent(), null, null);
-        setLanguageMatched(language, token.location);
-        return true;
     }
 
     @Override
     public boolean match_TagLine(Token token) {
-        if (!token.line.startsWith(TAG_PREFIX_CHAR)) {
+        if (token.line.startsWith(TAG_PREFIX_CHAR)) {
```
mpkorstanje, line 141: I intentionally rewrote these to early-out style. It avoids indentation, makes the code a bit easier to read.

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
```diff
@@ -31,6 +31,11 @@ final class GherkinTokenMatcher implements TokenMatcher {
     private KeywordMatcher currentKeywordMatcher;
     private String activeDocStringSeparator = null;
     private int indentToRemove = 0;
+    private Function<Line, KeywordMatcher.Match> matchFeatureKeyword;
+    private Function<Line, KeywordMatcher.Match> matchBackgroundKeyword;
+    private Function<Line, KeywordMatcher.Match> matchRuleKeyword;
+    private Function<Line, KeywordMatcher.Match> matchScenarioKeyword;
+    private Function<Line, KeywordMatcher.Match> matchExamplesKeyword;
```
mpkorstanje, line 38: Good one.

mpkorstanje:
I'm not seeing the improvements reflected in JMH.

On `main`

```
Benchmark                          Mode  Cnt     Score   Error  Units
GherkinParserBenchmarkTest.parse  thrpt    5  2247.780 ± 8.826  ops/s
```

On `minor_improvements_after_unrolling`

```
GherkinParserBenchmarkTest.parse  thrpt    5  2186.411 ± 16.941  ops/s
```

java/src/main/java/io/cucumber/gherkin/GherkinTokenMatcher.java
mpkorstanje: Mmh. Even if I take just these changes, performance goes down a bit. 

```
GherkinParserBenchmarkTest.parse  thrpt    5  2177.097 ± 21.057  ops/s
```

java/src/main/java/io/cucumber/gherkin/Line.java
```diff
@@ -66,11 +66,4 @@ String substringTrimmed(int beginIndex) {
         return StringUtils.substringAndLeftTrim(text, beginIndex, textLength);
     }
 
-    boolean startsWithTitleKeyword(String keyword) {
-        int keywordLength = keyword.length();
-        return textLength > keywordLength &&
-               text.charAt(keywordLength) == TITLE_KEYWORD_SEPARATOR &&
-               text.startsWith(keyword);
```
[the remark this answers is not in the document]
jkronegg: Both function do the same but in a different order. The `startsWithTitleKeyword` looks for the ':' first, when the `startsWith` looks for the first letter. Out of technical Java technical details, this should make no difference. For example, "Rule:" and "Scenario:" differs both in their first letter and their position of the ':'.

From the JVM code (assuming all JVM have the same `String` implementation), we have:
- `String.charAt(int index)` is guarded by `index<this.length`
- `String.startsWith(String prefix)` is guarded by `prefix.length<=this.length`

Thus when non-matching keyword:
- `startsWithTitleKeyword`: computes 2 length and does 2 guards 
- `startsWith`: computes 2 length and does 2 guards

And when matching keyword:
- `startsWithTitleKeyword`: computes 4 length and does 3 guards 
- `startsWith`: computes 2 length and does 2 guards

So in theory the `startsWith` should be faster.

If we pass the pre-computed keyword length to `startsWithTitleKeyword`, this gives:
- when non-matching keyword: computes 1 length and does 2 guards
- when matching keyword: computes 3 length and does 2 guards
Then performance could be better.

I'll investigate...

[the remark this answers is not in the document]
mpkorstanje: Okay. Maybe the best thing would be to split this PR up. Your explanation sounds plausibele. So maybe one of the other aspects slows it down more?

jkronegg:
Closed as the proposed improvement are not really convincing. I'll try to improve other parts of the code.

PR #455, opened 2025-08-17
Hotfix Gherkin.csproj

Romfos:
### 🤔 What's changed?

- Remove version pinning

Version pining is an antipattern. It make impossible to do a hotfixes. also it create complexity in some cases when first part of code depends on older version of library, but the second depends on never, e.t.c

n\a

### 📋 Checklist:

n\a

----

### DIFF
--- a/dotnet/Gherkin/Gherkin.csproj
+++ b/dotnet/Gherkin/Gherkin.csproj
@@ -45,7 +45,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Cucumber.Messages" Version="[28.0, 29.0)"/>
+    <PackageReference Include="Cucumber.Messages" Version="28.1.0" />
   </ItemGroup>
 
 </Project>


Romfos:
I'm talking about problem like this:

image

mpkorstanje:
Where do you encounter this? And looking at Java the lower bound can probably go down to `19.1.4`.

mpkorstanje:
> Version pining is an antipattern. It make impossible to do a hotfixes. also it create complexity in some cases when first part of code depends on older version of library, but the second depends on never, e.t.c

Did you mean a version range?

In https://github.com/cucumber/gherkin/pull/426 I specifically asked for a range, because the messages project churns through majors quite quickly. Most of these don't impact Gherkin. With a version range we can release messages without also having to make releases for Gherkin.

Additionally, should messages get a major release that doesn't impact Gherkin, we can make feature releases for Gherkin by increasing the upper bounds.

Romfos:
> Where do you encounter this? And looking at Java the lower bound can probably go down to `19.1.4`.

It will not. Example of these errors: 
image

Romfos:
If you explicitly require version < 29 then version 30 cannot be consumed

mpkorstanje:
>> Where do you encounter this? And looking at Java the lower bound can probably go down to 19.1.4.
> It will not. Example of these errors:

Let me rephrase that. Which dependency graph is currently causing you problems?

And: Looking at Java I see that we have a version range from `19.1.4` inclusive to `29.0.0` exclusive. So for .Net we can potentially use a similar range.

> If you explicitly require version < 29 then version 30 cannot be consumed

As expected, which is why we'd then release a minor version of Gherkin.

gasparnagy:
I did a bit of research on this, and I think I understand now what @Romfos said.

## Problem with upper bounds

I think the problem is not really with the range, but with the upper bound. I checked a couple of packages and I couldn't find any with an upper bound restriction. So this is definitely not common/idiomatic in dotnet/nuget landscape.

Here are two examples of the kind of problem that can happen and I can imagine. The version resolution of nuget is sometimes pretty strange, so there can be other issues as well. 

### Scenario 1

Gherkin uses a subset of functionality from messages and only for specific use cases. For example with a simple parsing to the legacy AST it is not used at all.

Imagine you have a project that uses the legacy ast parsing from Gherkin, but use messages for something else. In this case they cannot update the messages dependency beyond the upper bound, even though there is no real conflict.

### Scenario 2

There is a new major version of messages and you want to check if it is compatible with Gherkin. You can't even try without making a homebrew version of it.

## Problem with lower bound

There is also a slight problem with setting the lower bound to 28.0 although there's already a 28.1 released. The NuGet resolution tends to choose the lowest dependent package from the range in some situations, so in this case 28.0. I don't know the exact circumstances but I've also run into that and it is pretty annoying.

## My recommendation

I'm not qualified to make a full recommendation but what I see is that most packages use a lower boundary (>=) dependency specification.

I know that it sounds strange, but the code @Romfos suggested is doing exactly that. I have no idea why the `Version=28.1.0` setting is compiled to a dependency >=28.1.0 but this is how nuget works.

So all in all I think we should go as @Romfos suggested, but these issues are not super critical at the moment, especially until we don't have messages v29. So we don't need urgency.

mpkorstanje:
Oef. That's some counter intuitive syntax and behaviour.

Does using `Version=28.1.0` also allow a version 29.0.0 to be used?
- If so how do we prevent Gherkin from pulling in a dependency that it can't use?
- If not, how do we ensure that the latest in a range of major versions is used?

In essence is it possible to write `Version=[19.*, 29.0.0)`?

PR #457, opened 2025-08-17
Standardise Python tooling configuration formats

kieran-ryan:
### 🤔 What's changed?

- Standardised Python tooling configuration file formats to TOML from both TOML and INI
- Bumped pytest above v6.1 and removed ‘junit_family’ configuration - as [the value is subsequently default](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=junit_family#confval-junit_family)
- Bumped coverage above v7.2 and replaced overrided exclude patterns config with a new key for extending default excludes
- Running pytest via coverage in test workflow

- Improve maintainability - only requiring to understand a single format and secondly makes configuration data types explicit; for instance, in the below configuration, the TOML expresses that ‘source’ is a list, whereas inspecting tooling documentation is required to see that it is a “List of packages of directories…” with INI compared to being misinterpreted as a single string value
  ```toml
  [tool.coverage.run]
  source = [“gherkin”]
  ```
  ```ini
  [run]
  source = gherkin
  ```
- TOML is conventional in Python (through standards such as [PEP-621](https://peps.python.org/pep-0621/)) and is default for config samples in tooling documentation (see [tox](https://tox.wiki/en/latest/config.html))
- Running pytest via coverage in the workflow ensures automated validation that coverage collection remains functional and that the pinned versions are compatible - even though we are currently not reporting the coverage itself and should revisit

NA

### 📋 Checklist:


mpkorstanje:
Cheers!

PR #459, opened 2025-08-29
Remove legacy Python input-output processor

kieran-ryan:
### 🤔 What's changed?

Removed the legacy input-output processor from the Python implementation - as is largely non-functioning and no longer fulfils the purpose in which it was conceived - assumed scoped to validate the implementation against internal test data.

- Clean up the Python gherkin parser implementation
- The processor has been broken for considerable time without any bug reports - the API provides very limited functionality and similarly for usage
- Printing pickles functionality is broken - incorrectly referencing Python's `compile` function rather than the parser's compiler API

  ```console
  (python) ➜  python (main) ✗ uv run check.py                                          
        Built gherkin-official @ file:///Users/user/cucumber/gherkin/python
  Installed 17 packages in 31ms
  Traceback (most recent call last):
    File "/Users/user/cucumber/gherkin/python/run.py", line 16, in <module>
      inout.process(input_stream, output_stream)
      ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/user/cucumber/gherkin/python/gherkin/inout.py", line 30, in process
      pickles = compile(gherkin_document, uri)
  TypeError: compile() missing required argument 'mode' (pos 3)
  ```

- When parser exceptions occur, unhandled exceptions are raised on attempting to print them - referencing a non-existent `error.message` attribute rather than casting the error message to a string

  ```console
           {
                ...
  >             "data": error.message,
                        ^^^^^^^^^^^^^
                "media": {
                    "encoding": "utf-8",
                    "type": "text/x.cucumber.stacktrace+plain",
                },
            }
  E           AttributeError: 'ParserException' object has no attribute 'message'
  ```

- Other observed issues include:
  - `print_source`, `print_ast`, `print_pickles` argumentsdo not provide default values; so must all be set - even if only one are two are of interest
  - Parser exception handling covers a large breadth of code rather than just the single line triggering the parsing
  - Lacking type hinting, so usage is ambiguous
  - Class attribute `stop_at_first_error` is not used and a source of disinformation

- 🏦 Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)

- Is there an equivalent representation in other parsers
- How does the intention of this module compare with the `streams` directory within the Python parser
- This module has been broken for a considerable period of time and is not referenced elsewhere within the parser - is there a case to be made for its removal - to a leaner parser implementation? Or is it worth fixing what's currently broken and reconsidering at a future point in time.

### 📋 Checklist:


kieran-ryan:
Thanks a mill for the guidance!

> So I would encourage you to strip out anything that is broken, unused and not covered by tests.

In that case, given the state of the released version of this module: _caught_ parser exceptions result in unhandled exceptions when the `print_errors` is called on them, printing pickles is broken (and would not be in a messages format in any case), and it only really prints the original source and AST when parsing succeeds; it may be pragmatic to remove this module altogether - given usage is very limited and unlikely. View it similarly or any differently, before I strip out the module?

PR #468, opened 2025-09-07
Improved StringUtils.isWhitespace performance

jkronegg:
### 🤔 What's changed?

Refactored `StringUtils.isWhitespace(char)` to improve performance.

The current implementation is based on Unicode whitespace definitions, which is good thing, but the performance is not very good. The idea was to move this Unicode whitespace definition to the test code, and have an optimized version in the library code. 

The proposed version is based on an array of `boolean` which tells whether an Unicode character is a whitespace or not.

Several variants have tested with JMH benchmark. The proposed variant is the one with the top performance, with a 3.3 times improvement over the original version (but also the one which consume the more memory: there's no free lunch). Details are provided in the section below.

Improving Gherkin parser performance.

There are 29 spaces in the Unicode 65536 characters, based on the character type and directionality, and the last space character is at index 12288. These characteristics are used in the the tested variants. Several variants have been tested:

- large array of boolean values
- `BitSet`
- the original method but with fast path for non-whitespace characters
- the original method but with fast path for whitespace and non-whitespace characters
- `HashSet` with fast path for non-whitespace chars
- condition-based
- the original method
- `HashSet`
- sequential searching in a char array
- binary search in a char array

There is the code for these variants:

```java
package io.cucumber.gherkin;


import java.util.*;

final class StringUtilsBenchmark {

    // the original implementation in Gherkin
    static boolean isWhitespace0(char c) {
        // Fast path for the common whitespace case
        return c == ' ' || c == '\t' || isCharacterTypeSpace(c) || isDirectionalitySpace(c);
    }

    // the original version with a better fast path for common non-whitespace characters
    static boolean isWhitespace1(char c) {
        // Fast path for the common whitespace and non-whitespace case
        return (c<=' ' || c>='\u0085') && (c == ' ' || c == '\t' || isCharacterTypeSpace(c) || isDirectionalitySpace(c));
    }

    static char[] WHITESPACES;
    static Set<Character> WHITESPACE_SET = new HashSet<>();
    static {
        StringBuilder sb = new StringBuilder();
        for (char c = 0; c < Character.MAX_VALUE; c++) {
            if (isCharacterTypeSpace(c) || isDirectionalitySpace(c)) {
                sb.append(c);
                WHITESPACE_SET.add(c);
            }
        }
        WHITESPACES = sb.toString().toCharArray();
    }

    // sequential search is an array of whitespaces
    static boolean isWhitespace2(char c) {
        for (char w : WHITESPACES) {
            if (c == w) {
                return true;
            }
        }
        return false;
    }

    // original version with a fast path for common non-whitespace characters, but no fast path for common whitespace characters
    static boolean isWhitespace3(char c) {
        // Fast path for the common non-whitespace case
        return (c<=' ' || c>='\u0085') && (isCharacterTypeSpace(c) || isDirectionalitySpace(c));
    }

    // O(1) search in a HashSet
    static boolean isWhitespace4(char c) {
        return WHITESPACE_SET.contains(c);
    }

    // O(1) search in a HashSet with a fast path for common non-whitespace characters
    static boolean isWhitespace5(char c) {
        return (c<=' ' || c>='\u0085') && WHITESPACE_SET.contains(c);
    }

    private static final int[] numbers = {
            9, 10, 11, 12, 13, 28, 29, 30, 31, 32,
            133, 160, 5760, 8192, 8193, 8194, 8195,
            8196, 8197, 8198, 8199, 8200, 8201, 8202,
            8232, 8233, 8239, 8287, 12288
    };

    static {
        // On s'assure que le tableau est trié
        Arrays.sort(numbers);
    }

    // binary search in a sorted array on whitespaces
    public static boolean isWhitespace6(char number) {
        // Recherche binaire
        return binarySearch(number) >= 0;
    }

    private static int binarySearch(int key) {
        int low = 0;
        int high = StringUtilsBenchmark.numbers.length - 1;

        while (low <= high) {
            int mid = low + (high - low) / 2;
            if (StringUtilsBenchmark.numbers[mid] == key) {
                return mid; // Nombre trouvé
            } else if (StringUtilsBenchmark.numbers[mid] < key) {
                low = mid + 1; // Cherche dans la moitié supérieure
            } else {
                high = mid - 1; // Cherche dans la moitié inférieure
            }
        }
        return -1; // Nombre absent
    }

    // memory footprint is 193 * 8 bytes = 1544 bytes
    private static final BitSet numberSet = new BitSet(12289); // Taille suffisante pour inclure 12288

    static {
        // Initialisation des bits correspondant aux nombres
        int[] numbers = {
                9, 10, 11, 12, 13, 28, 29, 30, 31, 32,
                133, 160, 5760, 8192, 8193, 8194, 8195,
                8196, 8197, 8198, 8199, 8200, 8201, 8202,
                8232, 8233, 8239, 8287, 12288
        };
        for (int number : numbers) {
            numberSet.set(number);
        }
    }

    // O(1) check using BitSet
    public static boolean isWhitespace7(char number) {
        return numberSet.get(number);
    }

    // memory footprint is 12289 * 1 byte = 12289 bytes
    private static final boolean[] numberPresence = new boolean[12289]; // Taille pour inclure 0 à 12288

    static {
        // Initialisation des valeurs connues
        int[] numbers = {
                9, 10, 11, 12, 13, 28, 29, 30, 31, 32,
                133, 160, 5760, 8192, 8193, 8194, 8195,
                8196, 8197, 8198, 8199, 8200, 8201, 8202,
                8232, 8233, 8239, 8287, 12288
        };
        for (int number : numbers) {
            numberPresence[number] = true;
        }
    }

    // O(1) check using a boolean array
    public static boolean isWhitespace8(char number) {
        // Vérifie directement dans le tableau
        return number < numberPresence.length && numberPresence[number];
    }

    // condition-based check
    public static boolean isWhitespace9(char c) {
        return ((c>=9) && (c<=13)) || ((c>=28) && (c<=32)) || (c==133) || (c==160) || (c==5760) || ((c>=8192) && (c<=8202)) || (c==8232) || (c==8233) || (c==8239) || (c==8287) || (c==12288);
    }

    private static boolean isCharacterTypeSpace(char c) {
        return (((
                (1 << Character.SPACE_SEPARATOR)
                        // Not in the definition, but a subset of isDirectionalitySpace
                        | (1 << Character.LINE_SEPARATOR)
                        // Not in the definition, but a subset of isDirectionalitySpace
                        | (1 << Character.PARAGRAPH_SEPARATOR)
        ) >> Character.getType(c)) & 1) != 0;
    }

    private static boolean isDirectionalitySpace(char c) {
        return (
                (((1 << Character.DIRECTIONALITY_WHITESPACE)
                        | (1 << Character.DIRECTIONALITY_PARAGRAPH_SEPARATOR)
                        | (1 << Character.DIRECTIONALITY_SEGMENT_SEPARATOR)
                ) >> Character.getDirectionality(c)) & 1) != 0;
    }

}

```

These variants are tested using a JMH benchmark over an array of chars containing `very_long.feature`, `emoji.feature` and the keywords for all Gherkin languages. All variants are tested over all 65536 Unicode characters against the original version to ensure they behave the same.

```java
package io.cucumber.gherkin;

import org.junit.jupiter.api.Test;

import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Paths;

import static org.junit.jupiter.api.Assertions.assertEquals;
@org.openjdk.jmh.annotations.Measurement(time = 1, timeUnit = java.util.concurrent.TimeUnit.SECONDS)
@org.openjdk.jmh.annotations.Warmup(time = 1, timeUnit = java.util.concurrent.TimeUnit.SECONDS)
public class StringUtilsBenchmarkTest {
    /**
     * A set of representative content to be used in benchmarks:
     * - very_long.feature
     * - i18n_emoji.feature
     * - all keywords from all languages
     */
    static char[] chars;
    {
        StringBuilder s = new StringBuilder();
        try {
            Files.lines(Paths.get("../testdata/good/very_long.feature")).forEach(s::append);
            Files.lines(Paths.get("../testdata/good/i18n_emoji.feature")).forEach(s::append);
            GherkinDialects.getDialects().forEach(dialect -> {
                dialect.getStepKeywords().forEach(s::append);
                dialect.getScenarioKeywords().forEach(s::append);
                dialect.getBackgroundKeywords().forEach(s::append);
                dialect.getExamplesKeywords().forEach(s::append);
                dialect.getFeatureKeywords().forEach(s::append);
                dialect.getRuleKeywords().forEach(s::append);
                dialect.getScenarioOutlineKeywords().forEach(s::append);
            });
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
        chars = s.toString().toCharArray();
    }

    @Benchmark
    public boolean isWhitespaceOriginal() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace0(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace1(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceSequentialSearchInCharArray() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace2(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceOriginalWithFastPathForNonWhitespaceChars() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace3(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceHashSet() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace4(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceHashSetWithFastPastForNonWhitespaceChars() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace5(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceBinarySearch() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace6(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceBitSet() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace7(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceLargeBooleanArray() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace8(chars[i]);
        }
        return hasWhitespace;
    }

    @Benchmark
    public boolean isWhitespaceConditionBased() {
        boolean hasWhitespace = false;
        for (int i = 0; i < chars.length; i++) {
            hasWhitespace |= StringUtilsBenchmark.isWhitespace9(chars[i]);
        }
        return hasWhitespace;
    }

    @Test
    void all_implementations_behave_the_same() {
        for (char c = Character.MIN_VALUE; c < Character.MAX_VALUE; c++) {
            boolean b0 = StringUtilsBenchmark.isWhitespace0(c);
            boolean b1 = StringUtilsBenchmark.isWhitespace1(c);
            boolean b2 = StringUtilsBenchmark.isWhitespace2(c);
            boolean b3 = StringUtilsBenchmark.isWhitespace3(c);
            boolean b4 = StringUtilsBenchmark.isWhitespace4(c);
            boolean b5 = StringUtilsBenchmark.isWhitespace5(c);
            boolean b6 = StringUtilsBenchmark.isWhitespace6(c);
            boolean b7 = StringUtilsBenchmark.isWhitespace7(c);
            boolean b8 = StringUtilsBenchmark.isWhitespace8(c);
            boolean b9 = StringUtilsBenchmark.isWhitespace9(c);
            assertEquals(b0, b1, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b2, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b3, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b4, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b5, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b6, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b7, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b8, "Mismatch for char " + (int) c + " '" + c + "'");
            assertEquals(b0, b9, "Mismatch for char " + (int) c + " '" + c + "'");
        }
    }

}
```

These are the results:

image

The fastest variant (large boolean array) is 3.3 times faster than the original version, but consumes 12 kbytes of memory.
The second fastest variant (`BitSet`) is 2.6 times faster than the original version and consumes 1.5 kbyte of memory.
In 2025, I think the memory costs less than the CPU, so I selected the fastest variant.

Out of the performance, I noticed that the recent introduction of Unicode character type and directionality added 4 spaces to the original 25 spaces, all belonging the Unicode category "B":

- FS (File Separator): Code point 28 (0x1C), used to separate files.
-  GS (Group Separator): Code point 29 (0x1D), used to separate groups of records.
- RS (Record Separator): Code point 30 (0x1E), used to separate records within a file.
- US (Unit Separator): Code point 31 (0x1F), used to separate units of data or fields.

These 4 "whitespaces" are generally used in binary files to separate groups of data, not in text files like Gherkin feature files, but I don't think it will be an issue.

### 📋 Checklist:


mpkorstanje:
As tradeoffs go, both `isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace` and `isWhitespaceOriginalWithFastPathForNonWhitespaceChars` would be acceptable without much discussion.

 But why is the variability in their benchmarks so high?

mpkorstanje:
The common use-case for `isWhitespace` is to test a sequence of whitespace characters followed by 1 non-space character.  So in a well formed feature file, `isWhitespace` should expect to test 4-8 spaces for every non-space character. 

The current benchmark design seems to be weighted heavily towards testing non-space characters - and that would influence the performance optimizations in the wrong direction.

jkronegg:
> As tradeoffs go, both `isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace` and `isWhitespaceOriginalWithFastPathForNonWhitespaceChars` would be acceptable without much discussion.
> 
> But why is the variability in their benchmarks so high?

I cannot explain the high variability for these two variants, but it's reproducible.

mpkorstanje:
Then without running anything, I reckon the distribution of the timings is bi-modal. One mode for characters on the fast path and one mode for characters on the slow path.

jkronegg:
> The common use-case for `isWhitespace` is to test a sequence of whitespace characters followed by 1 non-space character. So in a well formed feature file, `isWhitespace` should expect to test 4-8 spaces for every non-space character.
> 
> The current benchmark design seems to be weighted heavily towards testing non-space characters - and that would influence the performance optimizations in the wrong direction.

That's true, the benchmark was not reproducing the real conditions. Thus, I tested some different content:
- 8 standard whitespace (chr 32) + 'a': `s.append("        a");`
- 2 tabs (chr 9) + 'a': `s.append("\t\ta");`
- 8 standard whitespace (chr 32) + 'É' (an character outside the non-whitespace fast path): `s.append("        É");`
- 8 unconventional space (chr 10) + 'É' (a character outside the non-whitespace fast path): `s.append((char)10+(char)10+(char)10+(char)10+(char)10+(char)10+(char)10+(char)10+'É');`
- 1 unconventional space (chr 10) + 7 standard whitespace (chr 32) + 'a': `s.append((char)10+"       a");`

The benchmark results shows that:
- as long as we are using characters that are in the fast path (both whitespace and non-whitespace), `isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace` is the fastest method (1.6-1.8 times faster than the original method). In this case, `isWhitespaceLargeBooleanArray` is 1-3-1.6 times faster than the original method (=still faster than the original method, but not the fastest).
- as soon as we use characters that are not in the fast path (either whitespace or non-whitespace), `isWhitespaceLargeBooleanArray` is the fastest method (1.3-5.1 times faster than the original version). In this case, `isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace` is 0.7-1.2 times "faster": so yes, performance can be lower that the original method.

Thus:
- with `isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace` you get the best performance for Gherkin language with english character set
- with `isWhitespaceLargeBooleanArray` you get the best performance for Gherkin languages with non-english characters. 

So we could either select one method, or even use a method specific to the Gherkin language (although I'm not sure this worth the additional code complexity). What do you prefer ?

You'll find the extended benchmark results below:
image
image
image
image
image

jkronegg:
I tested another variant with the large boolean array and fast path for common non-whitespace and whitespace characters:
- when the characters to test are within the ones of the fast path: this is a fast as `isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace`. 
- when the characters to test are outsite the ones of the fast path: the performance is in between the one of `isWhitespaceOriginalWithFastPathWhitespaceAndNonWhitespace` and `isWhitespaceLargeBooleanArray`

To this variant would be the better of two worlds (but consumes 12 kbytes of memory).

mpkorstanje:
I came up with a faster-fast-path, it is competitive with `isWhitespaceLargeBooleanArray`.

```
    static boolean isWhitespace10(char c) {
        if (c == ' ' || c == '\t') {
            return true;
        }
        return (c < ' ' || c > '\u0085' - 1) && (isCharacterTypeSpace(c) || isDirectionalitySpace(c));
    }
```    

On my machine

mpkorstanje:
An no.  A switch case is comparable to the original method. 

```java
    static boolean isWhitespace11(char c) {
        // Fast path for the common case
        if (c == ' ' || c == '\t') {
            return true;
        }

        switch (c){
            case 10:
            case 11:
            case 12:
            case 13:
            case 28:
            case 29:
            case 30:
            case 31:
            case 133:
            case 160:
            case 5760:
            case 8192:
            case 8193:
            case 8194:
            case 8195:
            case 8196:
            case 8197:
            case 8198:
            case 8199:
            case 8200:
            case 8201:
            case 8202:
            case 8232:
            case 8233:
            case 8239:
            case 8287:
            case 12288:
                return true;
        }
        return false;
    }
```

jkronegg:
I made a lot of benchmarks with different variants, but could not reproduce the result you obtain with your[ latest variant](https://github.com/cucumber/gherkin/pull/468#issuecomment-3270830394) (in bold below):
image

However, when rewriting the method with pure boolean operators (without "if"), the performance becomes slightly higher your previous "ManualOptimizationRanges". Interestingly, the "noIf" variant of "ManualOptimizationRanges" does not improve the performance. It looks like the JVM is sometimes able to optimize and sometimes not (I'm on Open JDK 21). Regarding the unstable benchmark results I got in the past, I changed my machine settings with "Top performance battery settings" and the results are much better.

I [applied this idea to your latest touchups](https://github.com/cucumber/gherkin/pull/468/commits/0509713a6a7beb25afcdde802c03fff4708ff9ea) and got similar results.
The results looks correct because when comparing the two "noIf" variants, the implementation are very close:

    static boolean isWhitespaceNoIf(char c) {
        return (c == ' ' || c == '\t') ||
                (c < ' ' && isWhiteSpaceSlow(c)) ||
                (c >= '\u1680' && c <= '\u3000' && isWhiteSpaceSlow(c)) ||
                (c == '\u0085' || c == '\u00a0');
    }

    static boolean isWhitespaceLt32OrRange5760To12288ThenSparseSpacesNoIf(char c) {
        return (c == ' ' || c == '\t') // Cas commun
                || ((c < 32 || (c >= 5760 && c <= 12288))
                && (isWhiteSpaceSlow(c))) // plage basse et haute
                || (c == 133 || c == 160);// caractères spécifiques
    }

image


@mpkorstanje Can you run the benchmark on your machine to compare ?

mpkorstanje:
Interesting. I got:

```
Benchmark                                                                                   Mode  Cnt        Score       Error  Units
IsWhiteSpaceBenchmarkTest.benchmarkIsWhiteSpace                                            thrpt    5  1715732.879 ± 39129.375  ops/s
IsWhiteSpaceBenchmarkTest.benchmarkIsWhiteSpaceNoIf                                        thrpt    5  2242017.802 ± 27202.371  ops/s
IsWhiteSpaceBenchmarkTest.benchmarkIsWhiteSpaceSlow                                        thrpt    5   935411.412 ±  5808.657  ops/s
IsWhiteSpaceBenchmarkTest.benchmarkIsWhitespaceLt32OrRange5760To12288ThenSparseSpaces      thrpt    5  2253054.484 ±  7177.060  ops/s
IsWhiteSpaceBenchmarkTest.benchmarkIsWhitespaceLt32OrRange5760To12288ThenSparseSpacesNoIf  thrpt    5  2254269.776 ±  2685.825  ops/s
```

jkronegg:
That's fine for me as well. I have nothing to add to this PR (I have a few other things, but this will be in another PR).

mpkorstanje (commented):
LGTM, but I'm going to have a look at improving the documentation and comments. It's a bit jumbled at the moment.

PR #504, opened 2025-11-04
Upgrade the PHP static analysis to Psalm 6

stof:
### 🤔 What's changed?

This upgrades the tooling used for static analysis of the PHP code, and applies fixes for issues reported by the new version (which are all about cases where version 5 was not enforcing proper error handling in cases where PHP functions use a union type with either `false` or `null` returned on errors).

Version 5 of Psalm is not maintained anymore and is incompatible with the latest versions of PHP, preventing adding those versions in the CI setup.
I plan to add newer PHP versions in the CI setup as a follow-up.

I decided to add a dependency on the `composer/pcre` package. This is a small package (maintained by the composer team for usage in Composer itself) providing a type-safe API wrapper around the PCRE API. In particular, failures are reported using exceptions instead of returning `null` and letting you call `preg_last_error_msg()` to get info about the cause of the error.
If you prefer not adding such a dependency, I could add explicit error handling for the PHP API itself in each place using the regex. But it would basically duplicate part of that package, so it might not be worth it.

### 📋 Checklist:

### DIFF
--- a/php/bin/gherkin
+++ b/php/bin/gherkin
@@ -13,6 +13,8 @@ $options = ['predictable-ids', 'no-source', 'no-ast', 'no-pickles'];
 
 $selectedOptions = getopt('', $options, $restIndex);
 
+assert($selectedOptions !== false, "getopt returns false only when it cannot find the argv array.");
+
 $paths = array_slice($argv, $restIndex);
 
 // lazily-read list of Sources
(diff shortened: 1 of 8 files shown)


mpkorstanje:
I've requested a review from @ciaranmcnulty but if he doesn't get around to it by next week, please request one from me. PHP isn't my forte so it would be quicker if he reviews it.

>  If you prefer not adding such a dependency, I could add explicit error handling for the PHP API itself in each place using the regex. But it would basically duplicate part of that package, so it might not be worth it.

As a test framework it would be preferable to have no production dependencies. You could potentially lift somethings from https://github.com/cucumber/gherkin/pull/229. Though I'm not sure about the quality. :sweat_smile:

PR #507, opened 2025-11-09
java: Set release version to 17

mpkorstanje:
### 🤔 What's changed?

* Set default release version to 17
* Enable Error prone with Nullaway
* Enable Checkstyle

* https://github.com/cucumber/cucumber-jvm/issues/2962
* https://github.com/cucumber/cucumber-jvm/issues/3000

### 📋 Checklist:

- [x] Use `messages-ndjson`
- [x] https://github.com/cucumber/cucumber-parent/pull/62
- [x] https://github.com/cucumber/messages/pull/352
- [x] https://github.com/cucumber/messages/issues/282

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org).
 This document is formatted according to the principles of [Keep A CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Changed
+- [Java] Set release version to 17 ([#507](https://github.com/cucumber/gherkin/pull/507)).
 
 ## [37.0.1] - 2025-12-22
 ### Fixed


java/src/main/java/io/cucumber/gherkin/Token.java
```diff
@@ -2,24 +2,26 @@
...
+    @Nullable List<LineSpan> matchedItems;
     int matchedIndent;
-    String matchedLanguage;
-    StepKeywordType keywordType;
+    @Nullable String matchedLanguage;
+    @Nullable StepKeywordType keywordType;
     Location location;
 
-    private Token(Line line, Location location) {
+    private Token(@Nullable Line line, Location location) {
```
mpkorstanje: Ideally we have two different types of token here. Can we use records?

mpkorstanje: I don't think the EOF token is needed. If we have a careful look at `TokenScanner.read` we can probably replace it with `null`.

PR #599, opened 2026-05-27
Allow steps to contain both DataTable and DocString

kruczekrobert:
### 🤔 What's changed?

This change allows a single Gherkin step to contain both a DataTable and a DocString.

Previously, the grammar/parser treated `Step DataTable` and `Step DocString` as mutually exclusive alternatives. This updates the grammar and generated parsers so a step can include both argument types, and updates the AST/pickle compilation path across implementations to preserve both values.

Added testdata coverage for:

- a valid step containing both a DataTable and a DocString
- the existing invalid case where a step contains repeated DocStrings

Some step definitions can naturally need both structured tabular input and a larger text payload. Gherkin already supports both argument types independently, but the grammar currently prevents using them together on the same step.

This proposal makes that combination representable directly in Gherkin instead of requiring workarounds such as encoding one argument inside the other.

I would appreciate feedback on whether allowing both DataTable and DocString on a single step is acceptable from the Gherkin language/design perspective.

I would also like reviewers to check whether the AST and pickle representation chosen here is the right compatibility trade-off across the supported implementations.

### 📋 Checklist:


kruczekrobert:
I'll check what the problems are this week and fix them

kruczekrobert:
@mpkorstanje 

Unfortunately, after a little checking, apart from minor/cosmetic issues, I came to the conclusion that the change in elixir is a breaking change.

I think the current implementation changes the `Step` contract in a way that is breaking for Elixir. 

Today Elixir depends on `cucumber_messages` 32.x, where `GherkinDocument.Feature.Step` still has the old `argument` field. The PR updates the Elixir code to use `data_table` and `doc_string`, but those fields do not exist in the current Elixir messages struct, so the implementation fails at compile/type-check time.

A few possible ways forward:

1. Release/update `cucumber_messages` first

   Add `data_table` and `doc_string` to the messages schema/bindings while keeping the old `argument` field for compatibility. Then update Gherkin implementations to use the new fields.

   This is the cleanest long-term option, but it requires coordinating the messages release before this PR can pass in all languages.

2. Keep backward compatibility in language implementations

   Implement a compatibility bridge where possible:
   - if the messages binding supports `data_table` / `doc_string`, use them;
   - otherwise keep using the old `argument` field.

   This prevents compile-time breakage, but old bindings still cannot fully represent a step containing both a data table and a doc string. So this can only preserve old behavior, not provide full support for the new feature.

3. Scope the PR to parser support first

   Update the grammar/generated parsers to accept both arguments, but avoid changing the public messages representation until the messages schema supports it across implementations.

   This keeps the parser change smaller, but means the new AST/pickle output cannot be fully represented yet.

4. Treat this as a breaking change

   If the intended behavior is that `Step.argument` is replaced by independent `dataTable` and `docString` fields, then this should probably be handled as a breaking change with coordinated messages and implementation updates.

My preference would be option 1: update/release messages first, keep the old field deprecated for a transition period, and then update Gherkin implementations to emit both fields where supported. Otherwise we risk making some language implementations compile only against unreleased or incompatible message bindings.

It's not the easiest thing to try to implement something in all supported languages, we'll probably have to bounce around a bit, but there are no solved problems without hard work :)

mpkorstanje:
A breaking change is fine.

To facilitate ordering you'd need to update `cucumber/messages`. This requires adding a a constructor parameter, which is a breaking change in most languages we support. In turn, because Gherkin will depend on these new constructor parameters, the supported lower bound of messages goes up, which is also a breaking change. 

Additionally, I think going from either-or semantics to and/or semantics is a breaking change on its own. 

So adding yet another breaking change shouldn't be a problem.

> Today Elixir depends on cucumber_messages 32.x, where GherkinDocument.Feature.Step still has the old argument field. The PR updates the Elixir code to use data_table and doc_string, but those fields do not exist in the current Elixir messages struct, so the implementation fails at compile/type-check time.

Ultimately you're looking at https://github.com/cucumber/messages/issues/63. Elixir generates its messages of a protobuf schema rather than our hand written templates. Protobuf can actually properly express the either-or relationship.

I think a pragmatic approach would be to update the protobuf schema. I'm not too concerned about breaking changes here. If anyone using Elixir makes themselves known they can help out with https://github.com/cucumber/messages/issues/63. :smile: 

What does matter though is the JSON output. That shouldn't change in a backwards incompatible manner. But as far as I can tell, none of the proposed changes would break it.

mpkorstanje:
> I think a pragmatic approach would be to update the protobuf schema. I'm not too concerned about breaking changes here. If anyone using Elixir makes themselves known they can help out with https://github.com/cucumber/messages/issues/63. 😄

Nevermind that. Fixing the existing PR (https://github.com/cucumber/messages/pull/29) was trivial.

mpkorstanje:
And because https://github.com/cucumber/gherkin/pull/433 isn't included in Elixir yet either I wouldn't mind be okay taking care of the changes for Elixir.

kruczekrobert:
I made some corrections, can you run the test workflow again?

mpkorstanje:
> I made some corrections, can you run the test workflow again?

If you create the PR against your fork you can run the workflow there. That way you won't have to wait.

kruczekrobert:
@mpkorstanje 

I didn't know that - thanks for advice!

Everything should be fine now - you can re run it here also :)

kruczekrobert:
@mpkorstanje Who is responsible for reviewing these changes?

mpkorstanje:
No one. There is no duty beyond what we care to do. 

But in all likely hood I'll end up reviewing the changes.

kruczekrobert:
@mpkorstanje Any update?

mpkorstanje:
I'm waiting for you.

https://github.com/cucumber/common/issues/2315#issuecomment-4664795134

mpkorstanje:
I'm going to review and then merge this without including the argument index from https://github.com/cucumber/messages/pull/468. Looking at other ongoing PRs that is best done separately.

mpkorstanje (approved):
LGTM.

> I would also like reviewers to check whether the AST and pickle representation chosen here is the right compatibility trade-off across the supported implementations.

I've removed all traces of `argument` where it was used as union type. Because messages will get an upgrade this is effectively a breaking change anyway, so might as well break something else while we are  at it. 

> Today Elixir depends on cucumber_messages 32.x, where GherkinDocument.Feature.Step still has the old argument field. The PR updates the Elixir code to use data_table and doc_string, but those fields do not exist in the current Elixir messages struct, so the implementation fails at compile/type-check time.

I've upgraded Elixir to use https://github.com/cucumber/messages/issues/63 so that is all moot now.

kruczekrobert:
@mpkorstanje Beautiful! When can we expect a release?

I'll be going to godog to add support for this.

mpkorstanje:
Not until https://github.com/cucumber/gherkin/pull/643 and https://github.com/cucumber/gherkin/pull/633 are done. Feel free to help out with the former.

PR #606, opened 2026-06-06
Publish @cucumber/gherkin as dual ESM/CJS

aslakhellesoy:
## Motivation

`@cucumber/gherkin` is currently CommonJS-only — no `type: module`, no `exports` field, just a `main` pointing at a CJS `dist/index.js`. That made sense when most consumers were Node CJS, but the world's moved on:

- **Vitest** runs everything through an ESM loader. CJS-only deps get force-transformed and frequently misbehave around default-vs-named interop, top-level `await`, and dynamic `import()`. Several of our internal projects hit this when using `@cucumber/gherkin` under Vitest.
- **Cloudflare Workers / workerd** is ESM-only. A CJS-only package is effectively unusable there without a bundler step that knows how to unwrap CJS named exports — and even then results are inconsistent. We've been blocked from using gherkin directly in a worker.
- **Modern bundlers (Vite, Rollup, esbuild, Bun, Deno)** all prefer ESM. CJS-only packages can't be tree-shaken properly, they break SSR/edge frameworks like Next.js App Router and Nuxt 4, and several major projects have started flagging or downranking CJS-only deps.
- **`@cucumber/messages`** already ships dual ESM/CJS — having `gherkin` lag behind is awkward for the rest of the ecosystem.

The short version: keeping gherkin CJS-only is increasingly costly for users, and the fix is mechanical.

## What this does

Switches `@cucumber/gherkin` to a dual ESM/CJS build, modelled directly on the `@cucumber/messages` setup so the family stays consistent.

- `package.json` gets `type: module`, a proper `exports` map with `import`/`require` conditions, and keeps `main`/`module`/`types` for older resolvers.
- Two TypeScript projects: `tsconfig.build-esm.json` → `dist/esm/src/` (NodeNext ESM), `tsconfig.build-cjs.json` → `dist/cjs/src/` (CommonJS). A small `package.cjs.json` is copied into `dist/cjs/` so Node treats that subtree as CJS regardless of the top-level `type`.
- All relative imports in `src/` now carry explicit `.js` extensions (required by Node ESM + NodeNext resolution). `gherkin-javascript.razor` is updated too so regenerated `Parser.ts` keeps them.
- `src/gherkin-languages.json` is codegen'd into `src/gherkin-languages.ts` by a small `scripts/build-dialects.mjs`. This sidesteps the Node-version differences in JSON import attribute syntax (`assert { type: 'json' }` vs `with { type: 'json' }`) across the CI matrix.
- Removed a stray `module.exports = GherkinLine` at the bottom of `GherkinLine.ts` that was incompatible with ESM emit.
- Added `scripts/smoke-esm.mjs` and `scripts/smoke-cjs.cjs` that exercise the published artifacts from both module systems — wired into `npm test` so CI catches regressions.
- `mocha` now runs under the `ts-node/esm` loader.

## Release pipeline

Untouched. `polyglot-release` still just runs `npm version`. `release-npm.yml` still runs `npm install-ci-test && npm publish`, and `prepublishOnly` runs the new dual build. The published tarball now contains `dist/cjs/` and `dist/esm/` instead of a flat `dist/`.

## Test plan

- [x] `npm test` — 36 mocha tests pass under ts-node ESM, plus ESM + CJS smoke tests against the built artifacts
- [x] `make acceptance` — full feature-file diff suite passes
- [x] `npm run lint` — biome clean
- [x] External consumer check: `import { Parser } from '@cucumber/gherkin'` and `require('@cucumber/gherkin')` both resolve via the `exports` field from a separate `type: module` project
- [ ] CI matrix (Node 18/20/22/23 × ubuntu/windows/macos)

If this lands cleanly we can apply the same pattern to the other `@cucumber/*` packages that are still CJS-only.


mpkorstanje:
@aslakhellesoy thanks for the PR.

Unfortunately it looks like this PR description was written by generative AI without much, if any, human editing.

So far contributions made this way have been rather low quality and we're in the process of adopting some guidelines for that (see https://github.com/cucumber/.github/pull/72/changes).

For now it would be much appreciated if you could not use an LLM for personal communication.

I'd also like to note that we currently have https://github.com/cucumber/common/issues/2232 to track progress on making packages ESM only so this PR would do more than desired.

But for further technical aspects I'll defer to @davidjgoss.

davidjgoss:
Yep, straight to ESM-only is the way to go here. I'll start the cascade of those this week with messages and gherkin.

aslakhellesoy:
Sorry for the LLM message, that was lazy of me.

Do you want me to open a new PR with ESM fixes, or are one of you going to handle it?

davidjgoss:
I’ve got the ESM change in hand, will post back here when messages and gherkin are released.

davidjgoss:
@aslakhellesoy `messages`, `gherkin` and all the other libraries are now published as straight up ESM - let me know if you run into any issues.

aslakhellesoy:
Awesome - thanks a lot @davidjgoss!!

PR #617, opened 2026-06-14
[cpp] remove cmate use native cmake and cpm

daantimmer:
### 🤔 What's changed?

- Removed custom cmate scripts and replaced it with native cmake instead
- Use Cmake Package Manager (CPM) to (optionally) download nlohmann/json and cucumber/messages

The custom cmate script adds no value and only confusion on maintaining the cpp codebase. Secondly its an unmaintained script by a third party.

### 📋 Checklist:

----


mpkorstanje:
I've merged https://github.com/cucumber/gherkin/pull/643. You'll have to resolve some conflicts. Note that the argument index is emitted if-and-only-if both doc string and data table are used as arguments.

.github/workflows/test-cpp.yml
```diff
@@ -9,66 +9,154 @@ on:
...
         config:
-          - os: ubuntu-latest
-            cc: gcc
-          - os: ubuntu-latest
-            cc: clang
-          - os: macos-latest
-            cc: clang
-          - os: windows-latest
-            cc: cl
+          - { os: ubuntu-22.04, cxx: g++,      version: 10 }
```
luke-hill, line 35: Whilst these look reasonably unique. You could save a bit of time with a matrix then some custom includes. We do this in cucumber-rails.

https://github.com/cucumber/cucumber-rails/blob/main/.github/workflows/test-ruby.yml#L19

cpp/cmake/dependencies.cpm.cmake
```diff
@@ -0,0 +1,61 @@
+# renovate: datasource=github-tags packageName=nlohmann/json versioning=semver
+set(NLOHMANN_JSON_VERSION 3.12.0 CACHE STRING "Version of nlohmann_json to use")
+# renovate: datasource=github-tags packageName=cucumber/messages versioning=semver
```
mpkorstanje, line 3: @daantimmer if this works, it can be used any dependencies.cpm.cmake file.

mpkorstanje: Looks like it does. The dependencies show up in https://github.com/cucumber/gherkin/issues/60 under the regex manager.

@daantimmer the structure of this file for Gherkin is rather different. Can you make sure it looks the same as this file, then I can apply the magic comments.

daantimmer: Yes it's on my to do list to get things aligned as much as possible :-)

mpkorstanje: Never mind mind. It was trivial. https://github.com/cucumber/messages/commit/4247afb4db31d142095ba990fbb2a61cf0b4ab66

daantimmer: Sure is :-). Thanks for contributing. You are now a cmake expert 👍

Don't forget to add the same version to the find_package call though.

PR #630, opened 2026-06-26
javascript: Test without setting autocrlf

mpkorstanje:

### 📋 Checklist:


mpkorstanje:
Fails, as expected because biome defaults to `lf`. 

I would expect we should be able to use `auto` based on https://github.com/biomejs/biome/discussions/7598

PR #643, opened 2026-07-07
Emit datatable and docstring argument index

mpkorstanje:

Emitting the argument index allows Cucumber implementations to determine in which order the data table and doc string arguments should be provided.

Follow up on https://github.com/cucumber/gherkin/pull/599.

### 📋 Checklist:


go/pickles.go
```diff
@@ -225,17 +228,37 @@ func pickleStep(
...
+		pickleStepArgument := &messages.PickleStepArgument{
+			DocString: pickleDocString(docString, variableCells, valueCells),
+			DataTable: pickleDataTable(dataTable, variableCells, valueCells),
+		}
+		if docString.Location.Line > dataTable.Location.Line {
+			pickleStepArgument.DataTable.ArgumentIndex = int64(0)
+			pickleStepArgument.DocString.ArgumentIndex = int64(1)
+		} else {
+			pickleStepArgument.DataTable.ArgumentIndex = int64(1)
+			pickleStepArgument.DocString.ArgumentIndex = int64(0)
```
mpkorstanje: This doesn't work as expected. Because `ArgumentIndex` is optional, the json representation is `json:"column,omitempty"` which omits the value 0. Now the json representation don't match up with expectation.

Such a beautiful language.

mpkorstanje: Needs: https://github.com/cucumber/messages/commit/1762225bac2effabe97f4742cc387342c53f4562

PR #655, opened 2026-07-14
[cpp] only depend on/fetch cucumber_messages

daantimmer:
### 🤔 What's changed?

CPM's AddPackage now uses the git hash of cucumber/messages.
In addition, the explicit dependency on nlohmann/json has been removed. As there is a transitive dependency through cucumber/messages.

In general it is better to pin to hash than a git tag since git tags are mutable and hashes are not.
As for the transitive dependency, it only adds additional maintenance to keep the additional dependency. Since gherkin has a hard dependency on messages and messages on json it is safe to assume that when messages configures/builds that gherkin will also configure and build.

### 📋 Checklist:

----

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,18 +3,5 @@
   "extends": [
     "github>cucumber/renovate-config",
     "github>cucumber/renovate-config:messages-range-strategy-widen"
-  ],
-  "dependencyDashboard": true,
-  "customManagers": [
-    {
-      "customType": "regex",
-      "description": "Update version variables in dependencies.cpm.cmake",
-      "managerFilePatterns": [
-        "/(^|/)dependencies.cpm.cmake$/"
-      ],
-      "matchStrings": [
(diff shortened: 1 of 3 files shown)


.github/renovate.json
```diff
@@ -3,18 +3,5 @@
...
-    {
-      "customType": "regex",
-      "description": "Update version variables in dependencies.cpm.cmake",
-      "managerFilePatterns": [
-        "/(^|/)dependencies.cpm.cmake$/"
-      ],
-      "matchStrings": [
-        "# renovate: datasource=(?<datasource>[a-z-]+?)(?: depName=(?<depName>.+?))? packageName=(?<packageName>.+?)(?: versioning=(?<versioning>[a-z-]+?))?\\sset\\(.+? (?<currentValue>.+?) CACHE STRING \".+?\"\\)"
-      ]
-    }
```
daantimmer, line 18: Is the removal of this block really something that is required? @mpkorstanje

mpkorstanje: Yes. I've moved it to the global config. 

https://github.com/cucumber/renovate-config/blob/main/dependencies-cpm-cmake.json

PR #697, opened 2026-08-23
Fix double substitutions of example values in Ruby

raviqqe:
### 🤔 What's changed?

The Ruby implementation now substitutes scenario outline example values literally, using the block form of `String#gsub`, instead of passing them as replacement patterns.

Every other implementation already substitutes literally. This is Ruby catching up rather than new behavior.

The second argument of `String#gsub` is a replacement pattern, not a literal string. Because example values are data, any value holding a backslash before special characters is silently rewritten instead of inserted.

No, I don't think so.

### 📋 Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 ## [Unreleased]
 ### Fixed
 - (i18n) Remove trailing space from Texan Rule keyword.
-- [Ruby] Substitute example values into scenario outlines literally, as the other implementations
-  do, instead of reading them as replacement patterns.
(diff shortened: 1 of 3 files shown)


luke-hill:
Happy to approve and merge. I'm just curious is this a bug. If so what is the behaviour without this change? @raviqqe

luke-hill (approved):

raviqqe:
@luke-hill As in the unit tests, the old behaviour applies escape/substitution rules of the replacement string to the given example values, although this is pure string-to-string substitution.

This leads to weird behaviors like the following, assuming the placeholder is `<value>`:

- `\\` is injected as `\`. <- 😄 
- `\\\\` is also injected as `\`. <- 🤔 
- `\\\\\\\\` is injected as `\\`. <- 🤯 
- ``a`a`` is injected as `aa`. <- 🤯 
- `&` is injected as `<value>`. <- 🤯

PR #699, opened 2026-08-25
[Dart] Improve test coverage & CI enforcement

Fried-man:
### 🤔 What's changed?

The dart solution now has full line coverage via unit test by:
- excluding generated files from coverage reporting
- refactoring private methods to be public so they are testable
- adding more tests

Also, the dart CI now enforces full line coverage.

Outside of best practices, the low line coverage was brought up in [this issue](https://github.com/olexale/bdd_widget_test/issues/123).

N/A

### 📋 Checklist:

----


mpkorstanje (changes requested):
No. Please only writing meaning full tests and accept a coverage ratio below 100%.

I didn't review everything because you're using generative AI. Please extrapolate from these highlights:

* Don't use `testdata/{good,bad}`. That's for the acceptance tests, don't duplicate them.
* Don't duplicate functionality already covered by the acceptance tests. 
* `transformNode` shouldn't be made public just for testing, nor should it be be tested in isolation. Though I don't think it needs coverage at all based on the previous point.

Fried-man:
> accept a coverage ratio below 100%

What minimum coverage requirement would you prefer?

mpkorstanje:
It's up to you to work out what can be reasonably achieved without duplicating the acceptance tests. For reference, in Java, excluding the generated parser, I've got 95% of instructions covered and 83% of all branches. But you shouldn't expect similar numbers because the structure of Dart is meaningfully different.

Fried-man:
The public API surface is now completely unmodified. The behavior is the same too.

